PR #85, opened 2024-05-31
[Search] Template Search service with Docker config

### LINKED ISSUE #10: Set up base project for Search service
## Feature Description

Create a place for the Algo team to work on Search.

The algo team needs a place to easily start development of the search functionality.

MNThomson:
# Description

Created the template for the `search` service built with Python, FastAPI, and Docker.

Closes #10

## How to Test

```bash
docker-compose up --build

curl localhost:8221/api/search
# Should return "Hello World"
```

## Checklist


TobySMurray (approved):
LGTM

apps/algorithm/search/requirements.txt
GDeane, line 1: Do we not need elasticsearch here, or is that going to be defined elsewhere?

GDeane:
The how to test was inaccurate, should be

docker-compose up --build

curl localhost:8221/api/**search**
# Should return "Hello World"

GDeane (approved):
LGTM, minor question that I can ask later

apps/algorithm/search/requirements.txt
MNThomson: When the elasticsearch database is created, we will add an example call from the python code (so then it will be added as a required python module). No point in adding it right now since there's no elasticsearch db to connect to.

PR #87, opened 2024-05-31
[Frontend] Template Frontend service with Docker config

MNThomson:
# Description

Closes #XXX

## Checklist


TobySMurray (approved):
LGTM

apps/frontend/index.html
```diff
@@ -0,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Martletplace</title>
```
sejal-agarwal, line 6: Can we change `Martletplace` to `MartletPlace` to keep it consistent everywhere?

apps/frontend/src/App.tsx
```diff
@@ -0,0 +1,7 @@
+import "./App.css";
+
+function App() {
+  return <h1>Martletplace</h1>;
```
sejal-agarwal, line 4: Change the name to `MartletPlace` here too pls 🫣

sejal-agarwal: Not a big deal though, we can update this too

sejal-agarwal (approved):
Thanks Max!

apps/frontend/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "scripts": {
+    "start": "vite",
+    "build": "tsc && vite build",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "format": "prettier --write ./src",
+    "format:ci": "prettier --check ./src",
+    "test": "vitest",
+    "preview": "vite preview"
+  },
+  "dependencies": {
```
CaelumD25, line 15: This will dev and regular dependencies will need:
- axios
- react-router-dom
- cypress
- @cypress/code-coverage
- @mui/material 
- @emotion/react 
- @emotion/styled
- vite-plugin-istanbul

apps/frontend/src/App.test.tsx
```diff
@@ -0,0 +1,5 @@
+import { expect, test } from "vitest";
```
CaelumD25, line 1: We didn't really plan on using vitest, but it apperantly works fine with cypress installed so this is likely fine, however in ci we'll need to configure it to run our cypress tests

CaelumD25 (approved):

ChloeZacharias (approved):

apps/frontend/index.html
MNThomson: My assumption is that wouldn't be capitalized...no?
image

apps/frontend/package.json
MNThomson: Yep, those can be added when you start needing them. This PR is just to template out the most basic frontend possible (trying to keep PRs extremely small)

apps/frontend/index.html
sejal-agarwal: We had a discussion about it and decided `MartletPlace` wasmore readable. Ever since we've been writing it like that in our docs.

apps/frontend/src/App.test.tsx
MNThomson: `vitest` is the `vite` native way of writing unit tests. It looks like you can use Cypress for unit tests, so if you want we can switch to just using Cypress (instead of both). If wanted, that'll be an easy swap, so I'm going to go ahead and merge this PR and we can open a follow up ticket if required.

PR #109, opened 2024-06-03
[Datalayer] Add base project

kylestang:
# Description

*This PR is built on top of #100, make sure to merge that first*
Adds the datalayer base project and docker configuration.
Sets up pgbouncer to connect to the database, and builds a rust executable which is copied into the datalayer.

Closes #25 

## How to Test

```bash
docker compose up --build

psql -p 8300 -U martletplace -h localhost
```

## Checklist
- [x] The code includes tests if relevant
- [x] I have *actually* self-reviewed my changes and done QA

PR #111, opened 2024-06-03
[Datalayer] Set up rust project

### LINKED ISSUE #41: Set up data layer rust project
## Feature Description

Create the rust project base for the data layer

The rust project needs to exist

## Related External Resources

nah

kylestang:
# Description

*This PR is built on top of #109, make sure to merge that first*
Adds the rust project to the datalayer and modifies the docker container to run both pgbouncer and rust.

Closes #41

## How to Test

```bash
docker compose up --build

curl localhost:8301
```

## Checklist

### DIFF
--- a/apps/data/datalayer/supervisord.conf
+++ b/apps/data/datalayer/supervisord.conf
@@ -1,0 +1,18 @@
+[supervisord]
+nodaemon=true
+
+[program:datalayer]
+command=/datalayer
+autorestart=true
+stopasgroup=true
+stdout_logfile=/dev/fd/1
+stdout_logfile_maxbytes=0
+redirect_stderr=true
+
+[program:pgbouncer]
+command=/opt/bitnami/scripts/pgbouncer/entrypoint.sh /opt/bitnami/scripts/pgbouncer/run.sh
+autorestart=true
+stopasgroup=true
+stdout_logfile=/dev/fd/1
+stdout_logfile_maxbytes=0
+redirect_stderr=true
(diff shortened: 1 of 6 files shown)


apps/data/datalayer/supervisord.conf
```diff
@@ -0,0 +1,18 @@
+[supervisord]
+nodaemon=true
+
+[program:datalayer]
+command=/datalayer
+autorestart=true
```
MNThomson, line 6: Do we want it to autorestart? Usually if something in a container dies, we want the entire container to die. Now this is a multi binary container which mucks this up. Thoughts?

docker-compose.yml
```diff
@@ -16,7 +16,8 @@ services:
     restart: unless-stopped
     depends_on: []
     volumes: [ ./apps/backend/user/src:/app/src ]
-    environment: []
+    environment:
+      - DB_URL=postgres://martletplace:johuigytfrdcv@datalayer:8300/martletplace
```
MNThomson, line 20: nit: let's change these all to be `DB_ENDPOINT`

MNThomson (approved):
LGTM after question + nit addressed

apps/data/datalayer/supervisord.conf
kylestang: Either way works. I went with this since the containers restart unless stopped and this seemed like the closest setting. If one of these services dies the whole app goes down so even though the container doesn't die we'll know pretty quick

PR #122, opened 2024-06-05
Frontend113 add dependencies and set up cypress

### LINKED ISSUE #113: Add Dependencies and set up Cypress
## Description

Adding dependencies to frontend react, and setting up cypress

We need our tech stack added to the 

## Related External Resources

See  SRS and #78

CaelumD25:
# Description

- Added dependencies
  - axios
  - @mui/material 
  - @emotion/react 
  - @emotion/styled 
  - react-router
  - cypress(dev)
  - vite-plugin-istanbul(dev)
  - @cypress/code-coverage(dev)

Cypress and vite configured, code coverage now works.

Closes #113 
## How to Test

No code implemented, no testing required other than testing if cypress opens
To do that: `npx open cypress`

## Checklist


apps/frontend/.gitignore
```diff
@@ -22,3 +22,8 @@ dist-ssr
 *.njsproj
 *.sln
 *.sw?
+
+# Fontend
```
sejal-agarwal, line 26: Small typo lol, *Frontend

sejal-agarwal (approved):
Looks great, thank you!! 😁 Just a tiny comment 🫣

apps/frontend/package.json
```diff
@@ -9,19 +9,26 @@
...
   },
   "dependencies": {
+    "@emotion/react": "^11.11.4",
+    "@emotion/styled": "^11.11.5",
+    "@mui/material": "^5.15.19",
+    "axios": "^1.7.2",
     "react": "^18.2.0",
-    "react-dom": "^18.2.0"
+    "react-dom": "^18.2.0",
+    "react-router": "^6.23.1"
```
MNThomson, line 22: nit: any explicit reason to add these deps in with the Cypress PR? I know you've mentioned that you'll be using them explicitly but it's good practice to only add a dep once it's required (let's say we have to revert this PR, we don't also revert the added deps that other parts of the code require).

MNThomson (approved):
LGTM from an ops standpoint

apps/frontend/package.json
CaelumD25: We should be using all of this stuff nearly immediately, there's a bunch of config that needed to be handled for getting cypress to have code coverage metrics and for working with Vite. Since we're doing TDD we're going to need cypress for any additions we're doing.

PR #123, opened 2024-06-05
Sejal/frontend121 create plain pages for martletplace

### LINKED ISSUE #121: Create plain pages for MartletPlace
## Feature Description

Add all the different pages we'll need for MartletPlace so that we can all code on different pages.

sejal-agarwal:
# Description

Added separate pages for application. 

This PR is a quick and dirty template for the pages we'll need. We will need to add protected routes for the pages that require authentication. We may also need to have a `listingId` when editing or viewing a listing.

Closes #121 

## Checklist

### DIFF
--- a/apps/frontend/src/App.tsx
+++ b/apps/frontend/src/App.tsx
@@ -1,9 +1,4 @@
-import {
-  BrowserRouter as Router,
-  Route,
-  Routes,
-  Navigate,
-} from "react-router-dom";
+import { BrowserRouter as Router, Route, Routes } from "react-router-dom";
 import Login from "./pages/Login";
 import ForgotPassword from "./pages/ForgotPassword";
 import CreateAccount from "./pages/CreateAccount";
@@ -22,7 +17,7 @@ function App() {
         <Route path="/" element={<Homepage />} />
 
         {/* If full URL is `/user/:id` then it would show another users profile,
-        otherwise show the current users account. */}
+        otherwise current users profile is shown. */}
         <Route path="/user" element={<Account />} />
         <Route path="/user/login" element={<Login />} />
         <Route path="/user/resetpassword" element={<ForgotPassword />} />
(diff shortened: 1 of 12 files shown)


apps/frontend/src/App.tsx
```diff
@@ -1,7 +1,36 @@
...
+        <Route path="/" element={<Navigate to="/login" />} />
+        <Route path="/login" element={<Login />} />
+        <Route path="/forgot-password" element={<ForgotPassword />} />
+        <Route path="/create-account" element={<CreateAccount />} />
+        <Route path="/homepage" element={<Homepage />} />
+        <Route path="/create-listing" element={<CreateListing />} />
+        <Route path="/edit-listing" element={<EditListing />} />
+        <Route path="/account" element={<Account />} />
+        <Route path="/view-listing" element={<ViewListing />} />
+        <Route path="/messages" element={<Messages />} />
```
MNThomson, line 30: Not sure if this has already been discussed but I would recommend creating the urls to be somewhat similar to the api endpoints for consistency. Perhaps something along the lines of:
```suggestion
        // If not logged in `<Navigate to="/login" />`
        <Route path="/" element={<Homepage />}

       // If the full URL contained the `:id` like `/user/:id` then it would show another users profile. If no `:id` then show the current users account.
        <Route path="/user" element={<Account />} /> 
        <Route path="/user/login" element={<Login />} />
        <Route path="/user/signup" element={<CreateAccount />} />
        <Route path="/user/reset" element={<ForgotPassword />} />

        <Route path="/listing/new" element={<CreateListing />} />
         // The full URL would be `/listing/edit/:id`
        <Route path="/listing/edit" element={<EditListing />} />
         // The full URL would be `/listing/view/:id`
        <Route path="/listing/view" element={<ViewListing />} />

        <Route path="/messages" element={<Messages />} />
```

sejal-agarwal: Thanks Max, addressed your comment!

CaelumD25 (approved):
Looks good to me!

PR #127, opened 2024-06-06
Brent/backend64 setup database tables

### LINKED ISSUE #64: Set up Database Tables
## Feature Description
Initial creation of the database tables for each data entity

There's no database tables

## Related External Resources
https://docs.google.com/document/d/1DNqexMBf4MpYPCxWsT3plUDAjDSbdJwUzw-UyZX-vDs/edit#heading=h.vbv6udt3y8om Section 4.2

bmachado18:
# Description

Created the main database tables almost as specified in the data entities section of the SRS 1.0. These are created within the initdb.sql file, where the docker container is initializing the database on each built. I'm curious how to manage persistent storage.

Notes:
- merged reviews and ratings into one ratings tables where reviews (which are a text content) are optional while ratings are not (and default zero)
- any timestamps (not including updated/modified ones) are by default assigned to the  current time (so time of creation)
- deletions of users cascade to SearchHistory, UserPreferences, And RatingTable. (so essentially deleting users won't effect listings and messages)
- dummy data is not added in initdb.sql.
- These tables and attributes are subject to change, but they act as a starting point.

I'm not sure how Resops wants to handle persistant storage, I could've mounted something but I am unsure. It also feels weird to push a big storage mount like that.

Closes #64 

## How to Test
1. close all docker images:
```docker-compose down -v```

2. Run database image:
```docker-compose up -d database```

4. Access postgreSQL server through psql on host:
```psql -p 8310 -U martletplace -h localhost```

5. Enter the password (see the docker-compose.yml for pw, under database)

7. connect to marketplace and view all tables
```
\c martletplace
\dt
```

## Checklist

### CODE REVIEW

File: apps/data/database/initdb.sql [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+
+CREATE TABLE UserTable (
+    user_id SERIAL PRIMARY KEY,
+    username VARCHAR NOT NULL,
+    email VARCHAR NOT NULL,
+    password VARCHAR NOT NULL,
+    name VARCHAR,
+    bio TEXT,
+    profile_pic_url VARCHAR,
+    joining_date TIMESTAMP DEFAULT NOW(),
```
Scott-Kenning, line 18: if we specified joining_date in the api spec we can leave it as that, but otherwise this sould probably be renamed to createdAt. Also will need a verified field as was just discussed in the meeting

apps/data/database/initdb.sql
```diff
@@ -0,0 +1,68 @@
...
+    listings_sold INTEGER[],
+    listings_purchased INTEGER[]
+);
+
+CREATE TABLE ListingTable (
+    listing_id SERIAL PRIMARY KEY,
+    seller_id INTEGER REFERENCES UserTable(user_id),
+    title VARCHAR NOT NULL,
+    description VARCHAR NOT NULL,
```
Scott-Kenning, line 27: Description should probably be nullable

apps/data/database/initdb.sql
Scott-Kenning: or date_created to line up with the other tables

apps/data/database/initdb.sql
```diff
@@ -0,0 +1,68 @@
+CREATE TYPE STATUS_TYPE AS ENUM ('AVAILABLE', 'SOLD');
```
Scott-Kenning, line 1: this will need REMOVED added

apps/data/database/initdb.sql
```diff
@@ -0,0 +1,68 @@
...
+CREATE TABLE MessageTable (
+    message_id SERIAL PRIMARY KEY,
+    sender_id INTEGER REFERENCES UserTable(user_id),
+    receiver_id INTEGER REFERENCES UserTable(user_id),
+    listing_id INTEGER REFERENCES ListingTable(listing_id),
+    message_body TEXT NOT NULL,
+    date_created TIMESTAMP DEFAULT NOW()
+);
+
+CREATE TABLE RatingTable (
```
Scott-Kenning, line 45: I think we are going with Reviews with a rating attached rather than Ratings with a review attached for naming

apps/data/database/Dockerfile
```diff
@@ -1,6 +1,5 @@
 FROM postgres:16-alpine
 
 COPY ./postgresql.conf /etc/postgresql.conf
-COPY initdb.sql /docker-entrypoint-initdb.d/initdb.sql
```
MNThomson: Our base database image should contain the `initdb.sql`, let's keep this in

MNThomson (commented):
I'm not sure why your `initdb.sql` file was being ignored. Are you running `docker-compose up --build` (if you omit build, it wouldn't have included the `initdb.sql`)

bmachado18:
I forgot the build flag, I'll change it back.

I'm curious, why not also just copy the file in the image for the populatedb.sql?

MNThomson:
> I'm curious, why not also just copy the file in the image for the populatedb.sql?

In the production environment, we don't want the db to be populated with dev/testing data. Since the Dockerfile is a source of truth for all environments, we use the docker compose volume mounts to add development only files.

Scott-Kenning (approved):

MNThomson (approved):
Dog LGTM Keyboard

PR #144, opened 2024-06-08
Brent/backend65 listing endpoints

### LINKED ISSUE #65: Create GET/POST/PATCH/DELETE endpoints for listing
## Feature Description
Initial functional creation of all /listing endpoints for the listing microservice

The front end needs to create, retrieve, edit and delete listings

## Related External Resources
[](https://docs.google.com/document/d/1DNqexMBf4MpYPCxWsT3plUDAjDSbdJwUzw-UyZX-vDs/edit) section 4.4.1

bmachado18:
# Description
Implemented each API's testing and functionality for listing microservice. This firstly involved configuring Cypress to the correct directory structure, then repeatedly writing tests equivalence class tests for an unimplemented API route, and after implementing the route until the tests pass. These endpoints defined within [the swagger UI under listings](http://market.lebl.ca/openapi/#/Listings/editListing), and within [the API list in the SRS doc](https://docs.google.com/document/d/1DNqexMBf4MpYPCxWsT3plUDAjDSbdJwUzw-UyZX-vDs/edit#heading=h.vf9vwfh56ne8).

An additional module installed was pg-promise with the command ```npm install pg-promise```

Keeping as a draft PR until all routes have implementations and tests. 
- [x] GET /api/listing/{id}
- [x] POST /api/listing
- [x] DELETE /api/listing/{id}
- [x] PATCH /api/listing/{id}

Concerns:
- In the API Agreement, GET listing includes the reviews of the listing, including the review information. Is that really necessary for every listing, or should it just return a list of review_id's attached to the listing? YES, since this api is mainly used by frontend and they user info need it on the listing page.
- In the API Agreement, GET listing does not have a comment/message describing the error. Should this be updated?
- in API Agreement, the post listing (Create) returns the newly created data. This violates REST principles, a POST should create and a GET should retrieve, right? So the POST should just return a message indicating success or not.

Closes #65 

## How to Test
There might be a better way to do this, but:
1. Remove all docker images, and build the following images:
```
docker-compose down

docker-compose up --build -d listing database datalayer
```
3. Go into exec of the listing service within docker, and the test suite:

image

```npm test```

## Checklist

### CODE REVIEW

**Comment by bmachado18** [comment]
@UVicMartletplace/backend I added a new route for the GET /api/listings/{id} but since I am new to node and express I was hoping you could review this first route so I can reduce changes for the other routes if done incorrectly or if the file structure or directory is off.

 I still need to parse the data to the response specified in the API Agreement (and its tests accordingly) but otherwise please leave a comment and let me make the changes! 
 
 ALSO take a look at the concerns section of this draft PR and maybe we can chat about them. 

THANKS!

File: apps/backend/listing/src/index.ts [hunk 1]
```diff
@@ -13,10 +24,31 @@ app.use((err: Error, req: Request, res: Response, next: NextFunction) => {
...
-  res.send("Hello world");
-});
+// GET /api/listing/:id - Get a listing's details
+const getListingById = (req: Request, res: Response, next: NextFunction) => {
+  const { id } = req.params;
 
-app.listen(PORT, "0.0.0.0", () => {
-  console.log(`Server running at http://0.0.0.0:${PORT}`);
-});
+  db.oneOrNone('SELECT * FROM listings WHERE listing_id = $1', [id])
```
Scott-Kenning, line 31: This should prbably be an async / await

apps/data/database/initdb.sql
```diff
@@ -0,0 +1,91 @@
+CREATE FUNCTION trigger_update_modified()
```
Scott-Kenning, line 1: This shouldn't be in this PR. If it's waiting on the DB changes then it should be on a branch off of the DB set up branch, doing it like this leads to conflicts if the DB branch is changed

Scott-Kenning: Same goes for all the other DB setup related files

bmachado18: I branched off the db setup, but I accidently made the PR to merge into master. my bad.

bmachado18:
Implemented each route as specified in the API spec. Other notes:
- We gave up on Jest due to transpilation errors; instead we are using Cypress. Currently these tests are mainly integration between database and listing service. I tried to use mocha and chai for unit tests and mocking but went into a rabbithole and gave up on that.
- In the database, added a distance attribute for listing; we talked about this not being used, in the past, and currently the createListing/updateListing API doesn't support modifying this distance, but it was needed so it complies with the listing GET request from the API spec
- CI is not working - I tried installing all the dependencies for cypress in the listings dockerfile, but then I get another error ([see these tests for details](https://github.com/UVicMartletplace/martletplace/pull/144/commits/19ffbf93ec2a32c84702eed39a6dcd0cc3ba4ecf)). I didn't want to get to deep into it.
- Auth is still not supported, so left a comment and we can make a new PR when it works.

apps/backend/listing/src/createListing.ts
```diff
@@ -0,0 +1,80 @@
...
+  const userID = 1;
+  const { listing } = req.body;
+
+  if (!listing) {
+    return res.status(400).json({ error: "missing parameter in request" });
+  }
+
+  const { title, description, price, location, images } = listing;
+
+  if (!title || !description || !price || !location || !images) {
```
Scott-Kenning, line 21: These should be combined into 1 error

apps/backend/listing/src/createListing.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const responseListing = {
+      listingID: createdListing.listing_id,
+      seller_profile: sellerProfile,
+      title: createdListing.title,
+      description: createdListing.description,
+      price: createdListing.price,
+      location: {
+        latitude: parseFloat(
+          createdListing.location.split(",")[0].replace("(", ""),
```
Scott-Kenning, line 61: this formatting should be different or skipped altogether since we should have the lat/long variables already

apps/backend/listing/src/index.ts
```diff
@@ -1,25 +1,49 @@
...
 
 const app = express();
+
 app.use(morgan("dev"));
 app.use(express.json());
 
+const pgp = pgPromise();
+const DB_ENDPOINT = process.env.DB_ENDPOINT || "";
+
+if (!DB_ENDPOINT && process.env.NODE_ENV !== "test") {
```
Scott-Kenning, line 19: why is the env being checked here? It should either not be checked or the error message should reflect the conditions being checked

apps/backend/listing/src/index.ts
```diff
@@ -1,25 +1,49 @@
...
+  updateListing(req, res, next, db),
+);
+app.delete("/api/listing/:id", (req, res, next) =>
+  deleteListing(req, res, next, db),
+);
 
-app.listen(PORT, "0.0.0.0", () => {
-  console.log(`Server running at http://0.0.0.0:${PORT}`);
-});
+if (process.env.NODE_ENV !== "test") {
```
Scott-Kenning, line 43: again, why are we checking the test env?

apps/data/database/initdb.sql
```diff
@@ -39,28 +36,27 @@ CREATE TABLE listings (
     location LOCATION_TYPE NOT NULL,
     status STATUS_TYPE NOT NULL,
     description VARCHAR,
-    image_urls TEXT[],
+    image_urls TEXT [],
     created_at TIMESTAMP NOT NULL DEFAULT NOW(),
-    modified_at TIMESTAMP NOT NULL DEFAULT NOW()
+    modified_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    distance FLOAT DEFAULT 5.0
```
Scott-Kenning, line 42: why is distance being added? Distance should be calculated relative to where the user sends the request from, so it won't be a consistent value in the DB

apps/data/database/populatedb.sql
```diff
@@ -5,12 +5,12 @@ INSERT INTO users (username, email, password, name, bio, profile_pic_url, verifi
...
-(2, 3, 'Listing Two', 'Description for listing two', 200, ROW(34.0522, -118.2437), 'SOLD', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Lucy']),
-(3, 4, 'Listing Three', 'Description for listing three', 300, ROW(41.8781, -87.6298), 'REMOVED', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Angel']),
-(4, NULL, 'Listing Four', 'Description for listing four', 400, ROW(29.7604, -95.3698), 'AVAILABLE', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Casper']),
-(5, NULL, 'Listing Five', 'Description for listing five', 500, ROW(51.5074, -0.1278), 'AVAILABLE', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Kiki']);
+INSERT INTO listings (seller_id, buyer_id, title, description, price, location, status, image_urls, distance) VALUES
+(1, NULL, 'Listing One', 'Description for listing one', 100, ROW(40.7128, -74.0060), 'AVAILABLE', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Jasper', 'https://api.dicebear.com/8.x/bottts/svg?seed=Bella', 'https://api.dicebear.com/8.x/bottts/svg?seed=Bella'], 5.0),
```
Scott-Kenning, line 9: Same thing about distance, since its not relative to a fixed point I don't think it makes sense to store it

apps/data/database/initdb.sql
bmachado18: OK I see, I just noticed it in the GET listing API so I added it. so I need to calculate the distance between the user location and the listing location, and send it with the get request?

Scott-Kenning: Yes that's correct, there might be some library that does lat/long calculations to use but I imagine it wouldn't be too difficult to implement as our own util function either, up to you which route you wanna take

bmachado18: OK I see, I just noticed it in the GET listing API so I added it. I'll remove that data attribute. Your right, it doesn't make much sense. Instead, would the logic be to take the user_id from the auth token - query and get user location. Get the listing location, then calculate the distance between them?

bmachado18: Users don't have a location. do we update the user table to store a location, or do we get their location on front-end and pass it in either in the header or through the request body, maybe as "user_location". request body below:
image

Scott-Kenning: user location will come in the request body

Scott-Kenning (approved):

apps/backend/listing/src/deleteListing.ts
```diff
@@ -0,0 +1,35 @@
...
+      [id],
+    );
+
+    if (result.rowCount === 0) {
+      return res.status(404).json({ error: "Listing not found" });
+    }
+
+    return res.status(200).json({ message: "Listing deleted successfully" });
+  } catch (err) {
+    return res.status(500).json({ error: "Something went wrong" });
```
TobySMurray, line 31: Not blocking but it would be good to kick back the error somewhere so we know what went wrong

TobySMurray (approved):
LGTM

apps/backend/listing/src/getListingById.ts
```diff
@@ -0,0 +1,121 @@
...
+        l.listing_id = $1
+    `;
+
+    const listing = await db.oneOrNone(listingQuery, [id]);
+
+    if (!listing) {
+      return res.status(404).json({ error: "Listing not found" });
+    }
+
+    console.log(listing.location);
```
MNThomson, line 56: Is this still left over from development?

apps/backend/listing/src/getListingById.ts
```diff
@@ -0,0 +1,121 @@
...
+    const listingLocationStr = listing.location;
+
+    const [latitudeStr, longitudeStr] = listingLocationStr
```
MNThomson, line 59:
```suggestion
    const [latitudeStr, longitudeStr] = listing.location
```

apps/backend/listing/src/getListingById.ts
```diff
@@ -0,0 +1,121 @@
...
+    `;
+
+    const imagesQuery = `
+      SELECT 
+        unnest(l.image_urls) AS "url"
+      FROM 
+        listings l
+      WHERE 
+        l.listing_id = $1
+    `;
```
MNThomson, line 104: Can't we get the url from the original SQL query?

apps/backend/listing/src/index.ts
```diff
@@ -1,25 +1,42 @@
...
 
 const PORT = 8212;
 
 const app = express();
+
 app.use(morgan("dev"));
 app.use(express.json());
 
+const pgp = pgPromise();
+const DB_ENDPOINT = process.env.DB_ENDPOINT || "";
```
MNThomson, line 17: This should definitely throw an error if not set. Defaulting to an empty string doesn't make sense

apps/backend/listing/src/updateListing.ts
```diff
@@ -0,0 +1,82 @@
...
+  try {
+    const updatedListing = await db.oneOrNone(
+      `UPDATE listings
+       SET title = $1,
+           description = $2,
+           price = $3,
+           location = $4,
+           image_urls = $5,
+           status = $6,
+           modified_at = NOW()
```
MNThomson, line 43: The database has triggers to automatically update the `modified_at` field

```suggestion
```

MNThomson (changes requested):
Anytime that an error is returned, can we add some sort of logging? It'll be really helpful when debugging the system if we can have a stack trace with the specific error and any additional context that would help

apps/backend/listing/src/getListingById.ts
bmachado18: You are right, I messed up. fixed it!

apps/backend/listing/src/deleteListing.ts
bmachado18: Added those changes, should've had those the whole time for dev

apps/backend/listing/src/index.ts
bmachado18: typescript throws an error so I'm just gonna leave it: /app/service/node_modules/ts-node/src/index.ts:859
    return new TSError(diagnosticText, diagnosticCodes, diagnostics);
           ^
TSError: ⨯ Unable to compile TypeScript:
src/index.ts(19,16): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string | IConnectionParameters<IClient>'.
  Type 'undefined' is not assignable to type 'string | IConnectionParameters<IClient>'.

    at createTSError (/app/service/node_modules/ts-node/src/index.ts:859:12)
    at reportTSError (/app/service/node_modules/ts-node/src/index.ts:863:19)
    at getOutput (/app/service/node_modules/ts-node/src/index.ts:1077:36)
    at Object.compile (/app/service/node_modules/ts-node/src/index.ts:1433:41)
    at Module.m._compile (/app/service/node_modules/ts-node/src/index.ts:1617:30)
    at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
    at Object.require.extensions.<computed> [as .ts] (/app/service/node_modules/ts-node/src/index.ts:1621:12)
    at Module.load (node:internal/modules/cjs/loader:1208:32)
    at Function.Module._load (node:internal/modules/cjs/loader:1024:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:174:12) {
  diagnosticCodes: [ 2345 ]
}

TobySMurray (approved):

MNThomson (approved):
Still not sure about using Cypress for unit/integration tests

bmachado18:
@MNThomson I understand you guys are getting graded on our tests - I dumped alot of time already trying to configure unit testing with mocha and chai alongside integration tests for cypress. We don't have any true unit tests so I can dump some more time to try and figure this out if you'd like.

Issue #77, opened 2024-05-31
𝗣𝗮𝗴𝗲: Create Search Page


CaelumD25:
## Feature Description

**View Listings: Users should be able to input their current location and view available items with pagination within a 5KM radius. They should be able to filter the items based on price and status (SOLD, AVAILABLE) and sort them by price, listed time and distance.**
This 

relates: #xxx
blocked by: #xxx

Issue #71, opened 2024-05-31
𝗣𝗮𝗴𝗲: Create Home/Main/Dashboard Page


CaelumD25:
## Feature Description

**View Listings: Users should be able to input their current location and view available items with pagination within a 5KM radius. They should be able to filter the items based on price and status (SOLD, AVAILABLE) and sort them by price, listed time and distance.**
This page should be created to facilitate users who have just been authenicated(from #69), it will contain a page of recommended items based on results from recommendation engine built by Algorithms. This will require components of: 
- #55 and
- #48 

_Note previously viewed listings will likely not be present on the main page as seen in the mockup below_

relates: #45, #42, #55, #48 
blocked by: #45, #42, #55, #48 

## Problem You're Trying to Solve

[See "See Personal Recommendations"](https://docs.google.com/document/d/1DNqexMBf4MpYPCxWsT3plUDAjDSbdJwUzw-UyZX-vDs/edit#heading=h.yl6jkxal18dy) and ["Indicate Disinterest"](https://docs.google.com/document/d/1DNqexMBf4MpYPCxWsT3plUDAjDSbdJwUzw-UyZX-vDs/edit#heading=h.h94p36c29gn1) from the SRS for User Stories

## Related External Resources

See [Get Recommendations](http://market.lebl.ca/openapi/#/Search%20and%20Recommendations/recommend) and [Listing Summary Object](http://market.lebl.ca/openapi/#/:~:text=Image-,ListingSummary,-LoginRequest) from the OpenAPI spec 

![Image](https://github.com/UVicMartletplace/martletplace/assets/54737724/c572efff-7bf5-4e5c-8956-fd7a49dc5536)

PR #153, opened 2024-06-12
Chloe/frontend/create search page #71 & #77

ChloeZacharias:
# Description

Created the home page and search page
image

image

image

Closes ##71 & #77 

## Checklist


apps/frontend/src/pages/Homepage.tsx
```diff
@@ -1,5 +1,246 @@
...
+  );
+
+  const handlePageChange = (
+    _event: React.ChangeEvent<unknown>,
+    page: number,
+  ) => {
+    setCurrentPage(page);
+  };
+
+  const handleSearch = () => {
```
CaelumD25, line 61: This will need to call the API as well, for example:

```js
axios.get({searchObject})
   .then((response) => console.log(response))
   .catch((error) => console.log(error))
```
can be mocked using 

```js
cy.intercept("GET", "/api/listing", {
  statusCode: 201,
  data: dataProvidedObject
})
```
If this is working off the spec, we should be able to change the axios url and everything should be plug and play

apps/frontend/src/components/filters.tsx
```diff
@@ -0,0 +1,215 @@
...
+  };
+
+  const handleTypeChange = (event: SelectChangeEvent<string>) => {
+    setType(event.target.value);
+  };
+
+  const handleApplyFilters = () => {
+    // Pass the current filter values to the parent component only when "Apply Filters" button is clicked
+    onFilterChange({
+      minPrice: minPrice === "" ? undefined : minPrice,
```
Scott-Kenning, line 68: Null should be used here over undefined. Undefined is usually reserved for a variable has been declared but has not yet been assigned a value

apps/frontend/src/components/listingCard.tsx
```diff
@@ -0,0 +1,110 @@
...
+});
+
+const convertDate = (dateString: string) => {
+  const dateTimeObject = new Date(dateString);
+  return dateTimeObject.toDateString();
+};
+
+const ListingCard: React.FC<ListingCardProps> = ({ listing }) => {
+  return (
+    <Grid
```
Scott-Kenning, line 31: I don't think Grid should be used here unless it's actually applying grid styling. I'm not super familiar with MUI, but I'd assume Box is the better choice

apps/frontend/src/components/listingCard.tsx
```diff
@@ -0,0 +1,110 @@
...
+      container
+      direction="column"
+      alignItems="flex-start"
+      justifyContent="flex-start"
+      spacing={3}
+      maxWidth={"29vw"}
```
Scott-Kenning, line 37: I haven't run the code, but I think that having this set at 29vw for width is going to mess up the mobile view. Ideally this would be 100% width and the spacing would be handled in the parent component, but if that's not simple to implement then a % width is still preferred over viewport units

apps/frontend/src/components/listingCard.tsx
```diff
@@ -0,0 +1,110 @@
...
+        item
+        xs={12}
+        sx={{
+          width: "90%",
+          maxHeight: "40vh",
+          display: "flex",
+          justifyContent: "center",
+          alignSelf: "center",
+          overflow: "hidden",
+          borderRadius: "8px 8px 8px 8px",
```
Scott-Kenning, line 50: borderRadius: "8px"

apps/frontend/src/components/listingCard.tsx
```diff
@@ -0,0 +1,110 @@
...
+        }}
+      >
+        <img
+          src={listing.imageUrl}
+          alt={listing.title}
+          style={{
+            width: "100%",
+            height: "auto",
+            objectFit: "cover",
+            borderRadius: "8px 8px 8px 8px",
```
Scott-Kenning, line 60: Same as last comment

apps/frontend/src/components/listingCard.tsx
Scott-Kenning: There's a few places where this occurs, but I'm just gonna leave this one comment for all of them

apps/frontend/src/components/listingCard.tsx
```diff
@@ -0,0 +1,110 @@
...
+        </Typography>
+        <Typography variant="body2" gutterBottom sx={{ color: "#616161" }}>
+          Posted: {convertDate(listing.dateCreated)}
+        </Typography>
+      </Grid>
+    </Grid>
+  );
+};
+
+ListingCard.propTypes = {
```
Scott-Kenning, line 97: This should be declared above the component

apps/frontend/src/components/searchBar.tsx
```diff
@@ -0,0 +1,252 @@
...
+const SearchBar: React.FC<SearchBarProps> = ({ onSearch, sortBy }) => {
+  const classes = useStyles();
+  const navigate = useNavigate();
+
+  const [showFilters, setShowFilters] = useState(false);
+  const [searchInput, setSearchInput] = useState("");
+  const [filters, setFilters] = useState<SearchObject>({
+    query: "",
+    minPrice: 0,
+    maxPrice: 100000,
```
Scott-Kenning, line 45: these max/min prices should either be null initially, or these values should be reflected in the filters component

apps/frontend/src/components/searchBar.tsx
```diff
@@ -0,0 +1,252 @@
...
+    setFilters((prevFilters) => ({ ...prevFilters, ...newFilters }));
+  };
+
+  const handleSearch = useCallback(() => {
+    const searchObject: SearchObject = {
+      ...filters,
+      query: searchInput,
+      sort: sortBy,
+    };
+    console.log("Search Object:", searchObject);
```
Scott-Kenning, line 73: console.log should be removed

apps/frontend/src/components/searchBar.tsx
```diff
@@ -0,0 +1,252 @@
...
+  const handleReload = () => {
+    navigate("/");
+    window.location.reload();
+  };
+
+  const handleListingRoute = () => {
+    navigate("/listing/view");
+  };
+
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
```
Scott-Kenning, line 107: This could probably be more easily implemented with just a showMenu toggle variable and css, but since this is already here there's no huge need to change it

apps/frontend/src/components/searchBar.tsx
```diff
@@ -0,0 +1,252 @@
...
+      justifyContent="center"
+      spacing={2}
+      sx={{
+        backgroundColor: "#f5f5f5",
+        boxShadow: "0px 4px 6px #808080",
+        paddingBottom: "10px",
+      }}
+    >
+      <Grid item>
+        {isSmallPage ? (
```
Scott-Kenning, line 133: `{isSmallPage &&
   <Box (or Grid maybe, not sure how MUI does it)>
     <Button> ... </Button>
   </Box>
 }
`

apps/frontend/src/components/searchBar.tsx
```diff
@@ -0,0 +1,252 @@
...
+              onClick={handleClick}
+              size="medium"
+              sx={{
+                ...classes.button,
+                fontSize: "40px",
+                maxHeight: "50px",
+                marginTop: "8px",
+                color: "white",
+              }}
+              aria-controls={open ? "account-menu" : undefined}
```
Scott-Kenning, line 224: null over undefined

apps/frontend/src/pages/Homepage.tsx
```diff
@@ -1,5 +1,246 @@
...
+    //FETCH API DATA??
+    // UPDATE LISTING OBJECTS
+    setSearchPerformed(true);
+  };
+
+  const fetchListings = async () => {
+    setListingObjects([
```
Scott-Kenning, line 68: this mock data object should be defined in a separate file

apps/frontend/src/pages/Homepage.tsx
```diff
@@ -1,5 +1,246 @@
...
+              <MenuItem value={"DISTANCE_DESC"}>Distance Descending</MenuItem>
+            </Select>
+          </FormControl>
+        </Box>
+      ) : (
+        <>
+          <Box
+            sx={{
+              borderBottom: "1px solid #808080",
+              fontSize: "1.3em",
```
Scott-Kenning, line 208: font sizes should be declared either in em or px, but not both. em is probably the better option but px seems to be what is used through most of the front end. Some devices will have different default font sizes, and em makes the font size relative to the parent element, so it can look weird on some screens if not carefully managed

apps/frontend/src/pages/Homepage.tsx
```diff
@@ -1,5 +1,246 @@
...
+        >
+          <Typography
+            variant="h5"
+            component="h3"
+            gutterBottom
+            sx={{ margin: "10px" }}
+          >
+            Results ({listingObjects.length})
+          </Typography>
+          <FormControl sx={{ minWidth: "20vw" }}>
```
Scott-Kenning, line 186: Same as previous comment about vw

apps/frontend/src/components/filters.tsx
```diff
@@ -0,0 +1,215 @@
...
+    setType("LISTING");
+    onFilterChange({
+      minPrice: undefined,
+      maxPrice: undefined,
+      status: "",
+      searchType: "",
+    });
+  };
+
+  const isDesktop = useMediaQuery("(min-width:850px)");
```
Scott-Kenning, line 88: I believe in the theme there are values defined for sm, md, lg and xl for screen sizes. These should be used for consistency if the theme variables are updated

apps/frontend/src/components/searchBar.tsx
```diff
@@ -0,0 +1,252 @@
...
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
+  const open = Boolean(anchorEl);
+  const handleClick = (event: React.MouseEvent<HTMLElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  const isDesktop = useMediaQuery("(min-width:1240px)");
```
Scott-Kenning, line 116: Same as above comment about theme screen sizes

apps/frontend/src/styles/pageStyles.tsx
```diff
@@ -10,27 +10,94 @@ export const useStyles = () => ({
     width: "100vw",
     height: "30%",
   },
+  HomePageBox: {
+    flexDirection: "column",
+    alignItems: "top",
+    justifyContent: "top",
+    width: "99.3vw",
```
Scott-Kenning, line 17: This is likely going to make the spacing funny or add a scroll bar on unusally sized screens. instead it should probably be set to 100vw, and the padding or margin should be removed elsewhere (setting `margin:0` and `padding:0` here might work)

Scott-Kenning (changes requested):
I went on the side of more comments rather than less, so not all of these need to be addressed. The ones about component width and mismatched types might break things, so those should be looked at before this gets merged.

It also wouldn't hurt to create a few more components to keep the files more organized, but again that's not a functional change so if you're in a rush it's not crucial

apps/frontend/src/components/listingCard.tsx
ChloeZacharias: I found this worked best when resizing to ensure that all the cards took up the same amount of space when resizing the page

apps/frontend/src/components/listingCard.tsx
ChloeZacharias: For this one specifically the grid worked best to make sure that all the information stayed in the same spot but there are some others that Ill change to boxes since I dont need the grid styling

apps/frontend/src/components/searchBar.tsx
ChloeZacharias: I needed to use undefined here or else the button caused errors

apps/frontend/src/components/filters.tsx
```diff
@@ -0,0 +1,215 @@
...
+  sort: string;
+  page: number;
+  limit: number;
+}
+
+interface FiltersProps {
+  onFilterChange: (filters: Partial<SearchObject>) => void;
+}
+
+const Filters: React.FC<FiltersProps> = ({ onFilterChange }) => {
```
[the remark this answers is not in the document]
ChloeZacharias: Updated

apps/frontend/src/components/filters.tsx
```diff
@@ -0,0 +1,215 @@
...
+  InputAdornment,
+  MenuItem,
+  SelectChangeEvent,
+  Select,
+  Typography,
+  FormHelperText,
+  useMediaQuery,
+} from "@mui/material";
+import { useStyles } from "../styles/pageStyles";
+import * as React from "react";
```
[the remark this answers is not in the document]
ChloeZacharias: updated

apps/frontend/src/components/filters.tsx
```diff
@@ -0,0 +1,215 @@
...
+import { colors } from "../styles/colors";
+import { ChangeEvent } from "react";
+
+interface SearchObject {
+  query: string;
+  minPrice: number;
+  maxPrice: number;
```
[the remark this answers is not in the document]
ChloeZacharias: Changed the types to match

apps/frontend/src/components/filters.tsx
```diff
@@ -0,0 +1,215 @@
...
+          fullWidth
+          sx={{
+            m: 1,
+            backgroundColor: colors.martletplaceNavyBlue,
+            "&:hover": { backgroundColor: colors.martletplaceBlueHover },
+            textTransform: "none",
+            fontSize: "16px",
+            padding: "13px 0",
+            align: "center",
+            margin: "22px 0px 0px 0px",
```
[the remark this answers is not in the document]
ChloeZacharias: updated

apps/frontend/src/components/filters.tsx
```diff
@@ -0,0 +1,215 @@
...
+        </Button>
+      </Box>
+
+      <Box
+        sx={{
+          width: "60%",
+          display: "flex",
+          flexDirection: "row",
+          gap: "10px",
+          marginBottom: "2vh",
```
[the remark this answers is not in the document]
ChloeZacharias: updated the margin to be in px

apps/frontend/src/components/filters.tsx
ChloeZacharias: Changed to null

apps/frontend/src/components/listingCard.tsx
ChloeZacharias: updated

apps/frontend/src/components/listingCard.tsx
ChloeZacharias: updated

apps/frontend/src/components/listingCard.tsx
ChloeZacharias: This would not let me declare it above the component so I did not change it

apps/frontend/src/components/searchBar.tsx
ChloeZacharias: changed the min/max prices to initially be null

apps/frontend/src/components/searchBar.tsx
ChloeZacharias: removed all the console.logs

apps/frontend/src/components/searchBar.tsx
ChloeZacharias: I kept it the same since its not causing any issues

apps/frontend/src/components/searchBar.tsx
ChloeZacharias: the pages for this need to be specific sizes because of how the buttons are styled so I did not update this

apps/frontend/src/components/searchBar.tsx
ChloeZacharias: Updated I kept the grid outside the if statement to keep the spacing consistent

apps/frontend/src/pages/Homepage.tsx
ChloeZacharias: changed to px

apps/frontend/src/components/filters.tsx
ChloeZacharias: I did not update this because I needed the page size that was checked to be specific

apps/frontend/src/styles/pageStyles.tsx
ChloeZacharias: updated

apps/frontend/src/pages/Homepage.tsx
ChloeZacharias: moved the mocked data to a different file

apps/frontend/src/pages/Homepage.tsx
ChloeZacharias: changed font-size to px

Issue #73, opened 2024-05-31
𝗣𝗮𝗴𝗲: Create "Create Listing" Page


CaelumD25:
## Feature Description

**Create/Manage Listings: Users should be able to create new listings, and be able to edit, delete existing listings**
This page is for creating a listing, with fields of title, description, price, and images; it will be copied for managing a listing but extra buttons will be added for editing and marking the status of an item. This page will be navigable from the account page -> "your listings" -> and a "Create Listing" option should be present at the bottom.

Handling the uploading of images will need help from @UVicMartletplace/backend so lets save this implementation for when their stuff is closer to being done as I don't really see how this functionality can be easily mocked.

This requires components of:
- #42 
- #45  

relates: #45, #42, #48 
blocked by: #45, #42 

## Problem You're Trying to Solve

[See SRS for User Story](https://docs.google.com/document/d/1DNqexMBf4MpYPCxWsT3plUDAjDSbdJwUzw-UyZX-vDs/edit#heading=h.xpqqo3a5mjt1)

## Related External Resources

[Creating a listing, see OpenAPI spec](http://market.lebl.ca/openapi/#/Listings/createListing) 
![Image](https://github.com/UVicMartletplace/martletplace/assets/54737724/6efbc99e-21d8-447e-b646-48dfd74ce0d3)

![Image](https://github.com/UVicMartletplace/martletplace/assets/54737724/372f5e65-81d2-40cd-9e54-02df8e2b5252)

Issue #72, opened 2024-05-31
𝗣𝗮𝗴𝗲: Create Individual Listings Page


CaelumD25:
## Feature Description

**View Individual Listing: Users should be able to open a listing and view the product card displaying attributes like images, title, description and current status.**
This page should be what is presented to users whenever they click on a listing summary card(see #48). It should at least include a title, description, price, status, and a image(can be placeholder for now). Below this listing reviews and review composition will eventually be available below the listing, however, it is not required for this sprint so that is of extremely low priority. This requires components of:
- #55 
- #51 

*Note: While there is a mocked page indicating the status of an item, it was rushed and I think we can do better to show that the item has been sold*

relates: #55, #51, #53, #52, #79 
blocked by: #55, #51

## Problem You're Trying to Solve

[See SRS for User Story](https://docs.google.com/document/d/1DNqexMBf4MpYPCxWsT3plUDAjDSbdJwUzw-UyZX-vDs/edit#heading=h.wjfgcg464hmc)

## Related External Resources

See [Get Listings](http://market.lebl.ca/openapi/#/Listings/getListing)
 and [The Listing Object](http://market.lebl.ca/openapi/#/:~:text=NewReview-,Listing,-NewListing) on the OpenAPI spec

![Image](https://github.com/UVicMartletplace/martletplace/assets/54737724/aacc3381-f5d2-49c4-a0ee-ab041c9c7617)
![Image](https://github.com/UVicMartletplace/martletplace/assets/54737724/40c47c5f-5501-4293-8912-2a1d1c5f0c03)

Issue #60, opened 2024-05-31
Create "Create Listing" Component


CaelumD25:
## Feature Description

This is the component that allows a user to create a listing
- Params:
  - NONE 

relates: #xxx
blocked by: #xxx

## Problem You're Trying to Solve

Develop a component that allows users to create a listing. See details of what the component should include in the screenshot below.

## Related External Resources

![Image](https://github.com/UVicMartletplace/martletplace/assets/72957712/65d20f64-77d1-4875-8a69-496fd39e499a)

Issue #51, opened 2024-05-31
Create View Individual Listing Component


CaelumD25:
## Feature Description

- Params:
  - JSON Object, Listing Object(OpenAPI spec)

relates: #xxx
blocked by: #xxx

## Problem You're Trying to Solve

When a user clicks on a particular listing, this is the view they will see. Note that the reviews component will be a separate ticket. 

## Related External Resources

![Image](https://github.com/UVicMartletplace/martletplace/assets/72957712/77e166b2-9538-4fbe-aa45-4981c9a6a39e)

PR #162, opened 2024-06-13
Caelum/frontend/72  create individual listings page

CaelumD25:
# Description

In this pull request implements the /listing/new and /listing/view/{id} routes. Editing a listing will follow shortly after this is approved as create listing and edit listing are essentially identical.

image

image

There may be an issue with CI where the headless cypress tests fail while the GUI ones do not.

Closes 
- #72
- #73 
- #60 
- #51 

## How to Test

Run `npx cypress run --component` to run tests headless, but the GUI works better it can be run by using `npx cypress open`

## Checklist


apps/frontend/src/App.tsx
```diff
@@ -30,11 +27,13 @@ function App() {
 
         <Route path="/listing/new" element={<CreateListing />} />
         {/* TODO: change full url to `/listing/edit/:id` */}
```
Scott-Kenning, line 29: these TODO comments can be removed now

apps/frontend/src/extra_components/Carousel.tsx
```diff
@@ -0,0 +1,139 @@
...
+  Paper,
+} from "@mui/material";
+import ArrowLeftIcon from "@mui/icons-material/ArrowLeft";
+import ArrowRightIcon from "@mui/icons-material/ArrowRight";
+
+export interface props {
+  imageURLs: string[];
+}
+const Carousel = ({ imageURLs }: props) => {
+  const [cards, setCards] = useState<React.ReactElement[]>([]);
```
Scott-Kenning, line 18: ReactElement should be imported at the top of the file along with useEffect, useState, etc.

apps/frontend/src/extra_components/Carousel.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const [currentPage, setCurrentPage] = useState(0);
+  const [slideDirection, setSlideDirection] = useState<
+    "right" | "left" | undefined
+  >("left");
+
+  const images: React.ReactElement[] = useMemo(() => {
+    return Array.from({ length: imageURLs.length }, (_, i) => <Card key={i} />);
```
Scott-Kenning, line 25: This can be simplified with array.map, but it's not super important

apps/frontend/src/extra_components/Carousel.tsx
```diff
@@ -0,0 +1,139 @@
...
+  }, [imageURLs]);
+
+  const handleNextPage = () => {
+    setSlideDirection("left");
```
Scott-Kenning, line 29: Since slideDirection is only ever left, it can probably just be hardcoded to left and the setSlideDirection logic can be removed

apps/frontend/src/pages/CreateListing.tsx
```diff
@@ -1,5 +1,281 @@
...
+      },
+      images: [],
+    },
+  });
+
+  // Gets the user location, and adds it to the listing object
+  const getUserLocation = useCallback(async () => {
+    try {
+      navigator.geolocation.getCurrentPosition((position) => {
+        newListingObject.listing.location.longitude = position.coords.longitude;
```
Scott-Kenning, line 59: I think this is modifying the properties of a const variable

apps/frontend/src/pages/CreateListing.tsx
```diff
@@ -1,5 +1,281 @@
...
+  }, [newListingObject.listing.location]);
+
+  // Updates and sends the newListingObject, to the server via post under /api/listing
+  const sendPostToCreateListing = async () => {
+    // In order to make sure that the images are retrieved before submitting
+    const successImages = await uploadImages();
+    let successLocation: boolean = true;
+    if (newListingObject.listing.location.latitude !== 0) {
+      successLocation = false;
+      getUserLocation().then(() => {
```
Scott-Kenning, line 75: I think this might set successLocation to true regardless of the return value from getUserLocation, has this been tested?

apps/frontend/src/pages/CreateListing.tsx
```diff
@@ -1,5 +1,281 @@
...
+      });
+    }
+    //console.log("Image Successful", imageUploadSuccessful)
+    console.log("New Listing Object", newListingObject);
+    if (successImages && successLocation) {
+      _axios_instance
+        .post("/listing", newListingObject)
+        .then((response) => {
+          alert("Listing Created!");
+          console.log("Response to upload", response);
```
Scott-Kenning, line 86: console.logs should be removed

apps/frontend/src/pages/CreateListing.tsx
```diff
@@ -1,5 +1,281 @@
...
+      );
+    }
+  };
+
+  // This function makes sure that the passed in url is a base64 data string
+  const isImageValid = (url: string) => {
+    try {
+      return url.startsWith("data");
+    } catch (error) {
+      return error === TypeError && url === undefined;
```
Scott-Kenning, line 107: If this is meant to return true when a url is undefined, that logic should be in the try block and not in the catch section along with an error. Returning true on an undefined url is also not ideal though because it's not a valid image, ideally intentionally undefined urls should be handled elsewhere in the code and this should return false on anything but a valid image url.

apps/frontend/src/pages/CreateListing.tsx
```diff
@@ -1,5 +1,281 @@
...
+            const response = await _axios_instance.post(
+              "/images",
+              { image: image },
+              {
+                headers: {
+                  "Content-Type": "multipart/form-data",
+                },
+              },
+            );
+            newListingObject.listing.images.push({ url: response.data.url });
```
Scott-Kenning, line 132: newListingObject should be updated with useState

apps/frontend/src/pages/CreateListing.tsx
```diff
@@ -1,5 +1,281 @@
...
+          } catch (error) {
+            console.error("Error uploading image:", error);
+          }
+        }),
+      );
+      return true;
```
Scott-Kenning, line 138: I believe with the way this is set up, any error uploading a specific image would be logged to the console, but the return would still be true. The outside catch would never see an error so the function would never return false, even if no image uploaded successfully.

apps/frontend/src/pages/CreateListing.tsx
```diff
@@ -1,5 +1,281 @@
...
+                <Grid item sm={4}></Grid>
+              </Grid>
+            </Grid>
+            <Grid item lg={6} xs={12}>
+              <Box>
+                <Typography variant={"h5"} sx={{ paddingLeft: "20px" }}>
+                  Image Preview
+                </Typography>
+                <Box sx={{ padding: "10px" }}>
+                  {!isImageValid(images[0]) ? (
```
Scott-Kenning, line 264: Since isImageValid currently returns true on undefined, this might not work as expected

apps/frontend/src/pages/ViewListing.tsx
```diff
@@ -1,5 +1,128 @@
...
+      .then((response) => {
+        setListingObject(response.data);
+      })
+      .catch((error) => {
+        console.log(error);
+      });
+  }, [id]);
+
+  // Convert price to string
+  const priceFormatter = new Intl.NumberFormat("en-CA", {
```
Scott-Kenning, line 50: This and convertDate are both identically defined here and in Chloe's search PR, it might be worth making a utils file for general shared functions like these

apps/frontend/src/pages/ViewListing.tsx
```diff
@@ -1,5 +1,128 @@
...
+  };
+
+  return (
+    <Container>
+      <Card sx={{ height: "100%", width: "100%" }}>
+        <CardContent>
+          <Grid container spacing={2}>
+            <Grid item sm={12} md={12} lg={6}>
+              <Typography variant={"h2"}>
+                {listingObject !== null
```
Scott-Kenning, line 73: `listingObject?.title ?? "Title not received"`

the ?? operator returns the left hand side unless it is null or undefined, in which case it returns the right hand side. This also misses the case where listingObject is defined but listingObject.title is null which could lead to an unhandled runtime error

apps/frontend/src/extra_components/Carousel.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <Grid
+        container
+        sx={{
+          display: "flex",
+          flexDirection: "row",
+          alignItems: "center",
+          alignContent: "center",
+          justifyContent: "center",
+          width: "100%",
+          height: "50vh",
```
Scott-Kenning, line 62: not a huge fan of using vh here, but it's not necessary to change as long as it displays properly on mobile and desktop

apps/frontend/src/extra_components/MultiFileUpload.tsx
```diff
@@ -0,0 +1,81 @@
...
+        };
+        reader.readAsDataURL(file);
+      }
+    }
+  };
+
+  return (
+    <Box>
+      {props.multipleUpload ? (
+        <input
```
Scott-Kenning, line 46: I think this could be simplified by setting `multiple={props.multipleUpload}` and then always rendering the same input. Also not sure why display is always set to none for style, but I will assume it works if this has been tested

Scott-Kenning (commented):
Looks pretty good, just a few small changes needed. There's one or two cases where errors aren't being handled properly and const objects are being modified incorrectly, everything else is less important to fix

apps/frontend/src/extra_components/Carousel.tsx
CaelumD25: Sounds good, will do!

apps/frontend/src/extra_components/Carousel.tsx
CaelumD25: I'll give that a shot, but it's working rn so I might not get around to it

apps/frontend/src/pages/CreateListing.tsx
CaelumD25: I tried modifying this with useState before, but despite it working for updating my newListingObject in other places it just doesn't work seeming randomly in other places

```ts
setNewListingObject((prevState) => ({
          ...prevState,
          listing: {
            ...prevState.listing,
            location: {
              ...prevState.listing.location,
              latitude = position.coords.latitude,
              longitude = position.coords.longitude,
            },
          },
        }));
```

apps/frontend/src/pages/CreateListing.tsx
CaelumD25: I've done quite a few manual tests, but I stopped when I realized that some browsers straight up block geolocation. I was thinking of changing this to instead default the location to the center of campus and then the other components can do the same if getting location fails.

It's not the most elegant solution, but there isn't much we can do without manual entry of a location

apps/frontend/src/pages/CreateListing.tsx
CaelumD25: Will do, wanted to wait until it had been reviewed before removing my "debugging" methods

apps/frontend/src/pages/CreateListing.tsx
CaelumD25: Yeah, I tried that for a long time, but it never seems to work, I had tried to reassign images to a list of `imageUrls = [{url: "www.example.com/image.png"},...] ` but it never actually updated the newListingObject. Is there another way to be updating that value, or maybe my code is just wrong?

```ts
setNewListingObject((prevState) => ({
          ...prevState,
          listing: {
            ...prevState.listing,
            images:  imageUrls
            },
          },
        }));
```

apps/frontend/src/pages/CreateListing.tsx
CaelumD25: Sounds good, I'll work on that today, there may be a couple things like that as I'm often working on this stuff quite late

apps/frontend/src/pages/ViewListing.tsx
CaelumD25: I like that idea, we're going to work on having a bit more consistency in today's meeting so I'll make sure to bring that up

apps/frontend/src/extra_components/Carousel.tsx
CaelumD25: Sounds good, I'll switch that to px or %, whichever works better!

apps/frontend/src/extra_components/MultiFileUpload.tsx
CaelumD25: Yeah... now that you say that you're definitely right, I'll squash that later today

apps/frontend/src/pages/CreateListing.tsx
Scott-Kenning: It's pretty important to use useState at all times when modifying a variable in state, it's probably the only thing in this PR that I would insist is fixed before approving. You could try something along the lines of

`setNewListingObject(
  {...newListingObject,
    listing: {
      ...newListingObject.listing,
      location: {
        <location stuff>
      }
    }
  }
)`

What you have there should work, but if it doesn't trying this might simplify it. If you're updating the object without using useState though, that would likely be the cause of errors when you do try and call useState

apps/frontend/src/pages/CreateListing.tsx
Scott-Kenning: That should work as far as I can tell. We can take a look at it together in the lab today

apps/frontend/tests/component_tests/CreateListing.cy.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+    // Attach image files (if this is necessary for your test)
+    cy.get("#image-input").attachFile([
+      "../../src/images/test_image1.jpg",
+      "../../src/images/test_image2.jpg",
+      "../../src/images/test_image3.jpg",
+    ]);
+
+    cy.get("#submit-button").click();
+
```
Scott-Kenning, line 125: adding a wait here might help

CaelumD25: Yeah, we tried adding waits everywhere but it still fails... So I've just changed the test for now and it just verifies the form data instead

CaelumD25: For some reason it has started working after I changed how the images are uploaded

apps/frontend/src/pages/ViewListing.tsx
CaelumD25: I'll make a seperate PR for this stuff later

apps/frontend/src/pages/CreateListing.tsx
CaelumD25: Hey Scott, if you could take another look at this code snippet, that'd be much appreciated!

Scott-Kenning: I'm still not 100% sure this works.

`try {`
`  for each image {`
`    try {`
`       (error thrown here on image upload failure)`
`    } catch {`
`       (error caught here)`
`    }`
`    // no error at this point since it was caught above`
`  }`
`} catch {`
`    // never triggered because error was already caught`
`}`

I would probably just have a single try/catch wrapping all the image uploads, since it is pretty unlikely that some would succeed and others would fail. If you want each upload to have its own catch and the promise to have its own as well, I think you can do a .catch() on the promise, or manually throw an error

https://stackoverflow.com/questions/24977516/catching-errors-in-javascript-promises-with-a-first-level-try-catch

CaelumD25: I think can see your point, but I think my brain is slowly dissolving...

Would this not work to make sure that all of the images have been uploaded if the user has a faulty connection with a high packet loss? I know it'd be cleaner to have the try catches attached to the axios promise, but I was trying that before and for whatever reason that weird test has started passing with this new implementation

ChloeZacharias (approved):

PR #201, opened 2024-06-23
Emila/70 search engine operations

### LINKED ISSUE #70: Implement Reindexing API
## Feature Description
Implement an API to reindex the search engine whenever there is an update in the backend database.

relates: #50 - These two should be tested together
blocked by: elasticsearch setup

Ensure the search index is updated with the latest listings information.

## Related External Resources
[5.1.4.2](https://docs.google.com/document/d/1DNqexMBf4MpYPCxWsT3plUDAjDSbdJwUzw-UyZX-vDs/edit#heading=h.mei2mmo2gmx2)

### Unit Tests

1. **Reindex Search Results When Listing Created:**
   - Test that the `reindexListingCreated` function processes a valid request payload correctly (creates the listing in the index).
   - Test that the `reindexListingCreated` function returns an appropriate success response.
   - Test that the `reindexListingCreated` function handles missing parameters in the request payload and returns an error response.
   - Test that the `reindexListingCreated` function correctly formats the listing data for reindexing.

2. **Reindex Search Results When Listing Edited:**
   - Test that the `reindexListingEdited` function processes a valid request payload correctly (edits the listing in the index).
   - Test that the `reindexListingEdited` function returns an appropriate success response.
   - Test that the `reindexListingEdited` function handles missing parameters in the request payload and returns an error response.
   - Test that the `reindexListingEdited` function correctly formats the edited listing data for reindexing.

3. **Reindex Search Results When Listing Deleted:**
   - Test that the `reindexListingDeleted` function processes a valid request payload correctly (deletes the listing from the index).
   - Test that the `reindexListingDeleted` function returns an appropriate success response.
   - Test that the `reindexListingDeleted` function handles missing parameters in the request payload and returns an error response.
   - Test that the `reindexListingDeleted` function correctly extracts the listing ID for reindexing.

### Integration Tests

1. **Backend to Search Engine Reindexing:**
   - Test that the backend correctly triggers the reindexing operation in the search engine when a listing is created.
   - Test that the backend correctly triggers the reindexing operation in the search engine when a listing is edited.
   - Test that the backend correctly triggers the reindexing operation in the search engine when a listing is deleted.
   - Test that the search engine index is updated correctly after each reindexing operation.

### Acceptance Tests

1. **Listing Created Reindexing:**
   - Given a new listing is created in the backend,
   - When the `/api/search/reindex/listing-created` endpoint is called with the correct payload,
   - Then the search engine index should be updated to include the new listing,
   - And the new listing should appear in search results.

2. **Listing Edited Reindexing:**
   - Given an existing listing is edited in the backend,
   - When the `/api/search/reindex/listing-edited` endpoint is called with the correct payload,
   - Then the search engine index should be updated with the edited listing details,
   - And the updated listing should appear in search results.

3. **Listing Deleted Reindexing:**
   - Given an existing listing is deleted in the backend,
   - When the `/api/search/reindex/listing-deleted` endpoint is called with the correct payload,
   - Then the search engine index should be updated to remove the deleted listing,
   - And the deleted listing should no longer appear in search results.

### Example Test Scenarios

1. **Scenario: Successful Reindexing on Listing Creation**
   - Given the backend creates a new listing,
   - When the reindexing API `/api/search/reindex/listing-created` is called with the correct payload,
   - Then the search engine index should be updated,
   - And the new listing should appear in search results.

2. **Scenario: Error Handling for Missing Parameters on Listing Creation**
   - Given the backend creates a new listing,
   - When the reindexing API `/api/search/reindex/listing-created` is called with a missing parameter,
   - Then the API should return an error response indicating the missing parameter.

3. **Scenario: Successful Reindexing on Listing Edit**
   - Given an existing listing is edited in the backend,
   - When the reindexing API `/api/search/reindex/listing-edited` is called with the correct payload,
   - Then the search engine index should be updated with the edited listing details,
   - And the updated listing should appear in search results.

4. **Scenario: Successful Reindexing on Listing Deletion**
   - Given an existing listing is deleted in the backend,
   - When the reindexing API `/api/search/reindex/listing-deleted` is called with the correct payload,
   - Then the search engine index should be updated to remove the deleted listing,
   - And the deleted listing should no longer appear in search results.

emilasuska:
# Description

This is a PR implementing search reindex API post, patch, and delete methods and their corresponding tests.

It includes completing:
1. All of these API endpoints with backend:
- `/api/search/reindex/listing-created`
- `/api/search/reindex/listing-edited`
- `/api/search/reindex/listing-deleted`

2. Integration tests using a test-index which is initialized before each test and destroyed after each test. All tests use the elasticsearch endpoint.

Closes #70 

## How to Test

1. Checkout this branch

4. In a terminal, navigate to the martletplace directory and run `docker-compose up --build`

5. Wait a minute for the data layer (particularly elasticsearch) to get ready for input

6. In a terminal, run `docker exec -it martletplace_search bash`

7. In the bash terminal, run `pytest src/test_server.py`

## How to manually test:

1. Uncomment this port in `docker-compose.yml`
image

2. Build the container

3. Wait a minute for the data layer (particularly elasticsearch) to get ready for input

4. In a browser, navigate to localhost:8221/docs

5. Play with the reindexing endpoints and search endpoint there by creating listings, searching for them based on different parameters, and deleting them

## Checklist
- [X] The code includes tests if relevant


apps/algorithm/search/requirements.txt
```diff
@@ -1,7 +1,9 @@
 fastapi
 pydantic
+# pydantic-partial
```
GDeane, line 3: Can you remove this if it's not being used?

apps/algorithm/search/src/test_server.py
```diff
@@ -1460,3 +1460,530 @@ def test_total_items_count_with_filter():
...
+        json=edited_listing_data,
+    )
+
+    assert response.status_code == 200
+    assert response.json() == {"message": "Listing edited successfully."}
+
+    es.indices.refresh(index=TEST_INDEX)
+    es_response = es.get(index=TEST_INDEX, id="test123")
+
+    assert es_response["found"] == True
```
GDeane, line 1737: I think this comparison to True is triggering ruff linting to be sad. Theus we should remove it (the next 2 lines effectively do the same thing)

apps/algorithm/search/src/test_server.py
```diff
@@ -1460,3 +1460,530 @@ def test_total_items_count_with_filter():
...
+        json=listing_data,
+    )
+
+    assert response.status_code == 200
+    assert response.json() == {"message": "Listing added successfully."}
+
+    es.indices.refresh(index=TEST_INDEX)
+    es_response = es.get(index=TEST_INDEX, id="test123")
+
+    assert es_response["found"] == True
```
GDeane, line 1491: Remove. It's making ruff linter sad

GDeane (approved):
LGTM, get CI/CD passing and maybe wait for comments from one more person.

I'm planning on doing a pretty significant refactoring of server.py right after this PR gets merged, so don't worry about that.

apps/algorithm/search/src/server.py
```diff
@@ -15,6 +15,18 @@
...
+        index=DEFAULT_INDEX,
+        body={
+            "mappings": {
+                "properties": {
+                    "location": {"type": "geo_point"},
+                }
+            }
+        },
+    )
+
```
MNThomson, line 29: Top level code is usually scary, usually only constants are up here. I guess having it here makes sense, but it still creeps me out a little

apps/algorithm/search/src/server.py
```diff
@@ -223,21 +249,30 @@ async def search(
 
 
 @app.post("/api/search/reindex/listing-created")
-async def reindex_listing_created(listingId: str, authorization: str = Header(None)):
-    # actual logic will go here
+async def post_listing(listing: Listing, authorization: str = Header(None)):
```
MNThomson, line 252: My assumption would be that fetching from the database would just be a better idea (since it's the source of truth).
```suggestion
async def post_listing(listingId: int, authorization: str = Header(None)):
```

MNThomson (commented):
These endpoints are publicly exposed and anyone on the internet can now arbitrarily modify listing data in the search database. What's the plan for security? My assumption is we don't want these endpoints exposed.

Also what happens if an error occurs in the backend before it sends the request to search? What if search is offline for a few minutes or is crashing? How do we ensure elasticsearch remains up to date with Postgres? There is [PGSync](https://pgsync.com/) ([GitHub](https://github.com/ankane/pgsync)) which might be interesting to investigate

apps/algorithm/search/src/server.py
GDeane: The Listing object defined here is different from the one defined in the database. We can update this in sprint 3 but for now it isn't broken so we're leaving it.

GDeane: Also this is following the definition laid out in the SRS

GDeane:
In terms of security. The user_id field is currently acting as a template which should be parsed from an authorization token (the method should fail if that isn't valid. That wasn't in our PR but we were assuming @MNThomson was going to add that when setting up the JWT token stuff.

As far as errors and PGSync, we're sticking to the working implementation based on our agreement for Sprint 2 to implement a minimum viable product, but for Sprint 3 it's a great idea.

Issue #91, opened 2024-06-01
User login and logout


Scott-Kenning:
## Feature Description
Implement the login and logout functionality, which involves creating the routes and implementing the token creation/expiry

blocked by: #90

## Problem You're Trying to Solve
Users must be able to logout / logout without having to create a new account every time

## Related External Resources
[section 4.4.1](https://docs.google.com/document/d/1DNqexMBf4MpYPCxWsT3plUDAjDSbdJwUzw-UyZX-vDs/edit#heading=h.vbv6udt3y8om)

PR #207, opened 2024-06-24
Backend user routes

### LINKED ISSUE #90: User Registration functionality
## Feature Description
Code for user registration, which involves creating a JWT token on sign up and returning it with the new user object to the front end

blocked by: #82

Users must be able to register for the website in order to use it

## Related External Resources
[section 4.4.1](https://docs.google.com/document/d/1DNqexMBf4MpYPCxWsT3plUDAjDSbdJwUzw-UyZX-vDs/edit#heading=h.vbv6udt3y8om)

Scott-Kenning:
# Description
All the backend user routes, including login/logout

Closes #90 #91

## How to Test
Run it

## Checklist

### DIFF
--- a/apps/backend/user/cypress.config.ts
+++ b/apps/backend/user/cypress.config.ts
@@ -1,0 +1,9 @@
+import { defineConfig } from "cypress";
+
+export default defineConfig({
+  e2e: {
+    baseUrl: 'http://localhost:8211',
+    specPattern: 'cypress/e2e/**/*.cy.{js,jsx,ts,tsx}',
+    supportFile: false,
+  },
+});
--- a/apps/backend/user/cypress/fixtures/example.json
+++ b/apps/backend/user/cypress/fixtures/example.json
@@ -1,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
--- a/apps/backend/user/src/createUser.ts
+++ b/apps/backend/user/src/createUser.ts
@@ -47,12 +47,12 @@ const createUser = async (
   const values = [username, email, hashedPassword, name, false];
 
   try {
-    const data = await db.oneOrNone(query, values)
-      if (!data) {
-        return res.status(500).json({ error: "User not created" });
-      }
+    const data = await db.oneOrNone(query, values);
+    if (!data) {
+      return res.status(500).json({ error: "User not created" });
+    }
 
-      return res.status(201).send(data);
+    return res.status(201).send(data);
   } catch (err) {
     console.log(err);
     return res.status(500).json({ error: "Something went wrong" });
--- a/apps/backend/user/src/createUser.ts
+++ b/apps/backend/user/src/createUser.ts
@@ -11,10 +11,10 @@ const createUser = async (
 ) => {
   const { username, password, email, name } = req.body;
 
-  if (!username || !password || !email) {
+  if (!username || !password || !email || !name) {
     return res
       .status(400)
-      .json({ error: "Username, password, and email are required" });
+      .json({ error: "Username, password, name and email are required" });
   }
 
   const hasDigit = /(?=.*\d)/.test(password);
@@ -40,20 +40,19 @@ const createUser = async (
 
   const query = `
     INSERT INTO users (username, email, password, name, verified)
-    VALUES ($1, $2, $3, $4, $5, )
+    VALUES ($1, $2, $3, $4, $5)
     RETURNING user_id, username, email, name, bio, profile_pic_url;
   `;
 
   const values = [username, email, hashedPassword, name, false];
 
   try {
-    await db.oneOrNone(query, values).then((data: User) => {
+    const data = await db.oneOrNone(query, values)
       if (!data) {
         return res.status(500).json({ error: "User not created" });
       }
 
       return res.status(201).send(data);
-    });
   } catch (err) {
     console.log(err);
     return res.status(500).json({ error: "Something went wrong" });
--- a/apps/backend/user/src/createUser.ts
+++ b/apps/backend/user/src/createUser.ts
@@ -9,7 +9,7 @@ const createUser = async (
   res: Response,
   db: IDatabase<object>,
 ) => {
-  const { username, password, email, name, bio, profile_pic_url } = req.body;
+  const { username, password, email, name  } = req.body;
 
   if (!username || !password || !email) {
     return res
@@ -39,9 +39,9 @@ const createUser = async (
   const hashedPassword = await bcrypt.hash(password, 10);
 
   const query = `
-    INSERT INTO users (username, email, password, name, bio, profile_pic_url, verified)
-    VALUES ($1, $2, $3, $4, $5, $6, $7)
-    RETURNING user_id, username, email, name, bio, profile_pic_url, verified, created_at, modified_at;
+    INSERT INTO users (username, email, password, name, verified)
+    VALUES ($1, $2, $3, $4, $5, )
+    RETURNING user_id, username, email, name, bio, profile_pic_url;
   `;
 
   const values = [
@@ -49,8 +49,6 @@ const createUser = async (
     email,
     hashedPassword,
     name,
-    bio,
-    profile_pic_url,
     false,
   ];
 
--- a/apps/backend/user/src/createUser.ts
+++ b/apps/backend/user/src/createUser.ts
@@ -9,7 +9,7 @@ const createUser = async (
   res: Response,
   db: IDatabase<object>,
 ) => {
-  const { username, password, email, name  } = req.body;
+  const { username, password, email, name } = req.body;
 
   if (!username || !password || !email) {
     return res
@@ -44,13 +44,7 @@ const createUser = async (
     RETURNING user_id, username, email, name, bio, profile_pic_url;
   `;
 
-  const values = [
-    username,
-    email,
-    hashedPassword,
-    name,
-    false,
-  ];
+  const values = [username, email, hashedPassword, name, false];
 
   try {
     await db.oneOrNone(query, values).then((data: User) => {
--- a/apps/backend/user/src/createUser.ts
+++ b/apps/backend/user/src/createUser.ts
@@ -1,0 +1,71 @@
+import { Request, Response } from "express";
+import { IDatabase } from "pg-promise";
+import { User } from "./models/user";
+import bcrypt from "bcryptjs";
+
+// createUser route
+const createUser = async (
+  req: Request,
+  res: Response,
+  db: IDatabase<object>,
+) => {
+  const { username, password, email, name, bio, profile_pic_url } = req.body;
+
+  if (!username || !password || !email) {
+    return res
+      .status(400)
+      .json({ error: "Username, password, and email are required" });
+  }
+
+  const hasDigit = /(?=.*\d)/.test(password);
+  const hasLowercase = /(?=.*[a-z])/.test(password);
+  const hasUppercase = /(?=.*[A-Z])/.test(password);
+  const hasSpecialChar = /(?=.*\W)/.test(password);
+  const hasMinLength = password.length >= 8;
+
+  if (
+    !hasDigit ||
+    !hasLowercase ||
+    !hasUppercase ||
+    !hasSpecialChar ||
+    !hasMinLength
+  ) {
+    return res
+      .status(400)
+      .json({ error: "Password does not meet constraints" });
+  }
+
+  // Hash the password
+  const hashedPassword = await bcrypt.hash(password, 10);
+
+  const query = `
+    INSERT INTO users (username, email, password, name, bio, profile_pic_url, verified)
+    VALUES ($1, $2, $3, $4, $5, $6, $7)
+    RETURNING user_id, username, email, name, bio, profile_pic_url, verified, created_at, modified_at;
+  `;
+
+  const values = [
+    username,
+    email,
+    hashedPassword,
+    name,
+    bio,
+    profile_pic_url,
+    false,
+  ];
+
+  try {
+    await db.oneOrNone(query, values).then((data: User) => {
+      if (!data) {
+        return res.status(500).json({ error: "User not created" });
+      }
+
+      return res.status(201).send(data);
+    });
+  } catch (err) {
+    console.log(err);
+    return res.status(500).json({ error: (err as Error).message });
+  }
+};
+
+export { createUser };
--- a/apps/backend/user/src/createUser.ts
+++ b/apps/backend/user/src/createUser.ts
@@ -64,7 +64,7 @@ const createUser = async (
     });
   } catch (err) {
     console.log(err);
-    return res.status(500).json({ error: (err as Error).message });
+    return res.status(500).json({ error: "Something went wrong" });
   }
 };
 
--- a/apps/backend/user/src/deleteUser.ts
+++ b/apps/backend/user/src/deleteUser.ts
@@ -57,7 +57,7 @@ const deleteUser = async (
     });
   } catch (err) {
     console.log(err);
-    return res.status(500).json({ error: (err as Error).message });
+    return res.status(500).json({ error: "Something went wrong" });
   }
 };
 
--- a/apps/backend/user/src/deleteUser.ts
+++ b/apps/backend/user/src/deleteUser.ts
@@ -1,0 +1,64 @@
+import { Request, Response } from "express";
+import { IDatabase } from "pg-promise";
+
+const deleteUser = async (
+  req: Request,
+  res: Response,
+  db: IDatabase<object>,
+) => {
+  // This is like this so resops can replace the userid but the guard on id === 1 doesn't show an error, I'm not crazy
+  const id = 1 === 1 ? 2 : 1;
+
+  if (id === 1) {
+    return res.status(401).json({ error: "Cannot delete the base user" });
+  }
+
+  const query = `
+  BEGIN;
+
+  -- Update messages table to display deleted user
+  UPDATE messages
+  SET sender_id = 1
+  WHERE sender_id = $1;
+
+  UPDATE messages
+  SET receiver_id = 1
+  WHERE receiver_id = $1;
+
+  -- Update reviews table to display deleted user
+  UPDATE reviews
+  SET user_id = 1
+  WHERE user_id = $1;
+
+  -- Update listings table buyer_id to deleted user
+  UPDATE listings
+  SET buyer_id = 1
+  WHERE buyer_id = $1;
+
+  COMMIT;
+`;
+
+  const query2 = `    
+-- Delete user
+  DELETE FROM users
+  WHERE user_id = $1
+  RETURNING *;    
+  `;
+
+  try {
+    await db.none(query, [id]);
+
+    await db.oneOrNone(query2, [id]).then((deletedUser) => {
+      if (!deletedUser) {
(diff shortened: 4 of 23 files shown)


TobySMurray (approved):
Lovely code, lgtm!

apps/backend/user/src/login.ts
```diff
@@ -0,0 +1,57 @@
...
+      userID: user.user_id,
+      username: user.username,
+      name: user.name,
+      email: user.email,
+      bio: user.bio,
+      profileUrl: user.profile_pic_url,
+    });
+  } catch (err) {
+    console.error(err);
+    return res.status(500).json({ error: (err as Error).message });
```
kylestang, line 53:
```suggestion
    return res.status(500).json({ error: "Something went wrong" });
```

apps/backend/user/src/patchUser.ts
```diff
@@ -0,0 +1,106 @@
...
+  const id = 1 === 1 ? 3 : 1;
+
+  const { username, password, name, bio, profilePictureUrl, verified } =
+    req.body;
+
+  if (!id) {
+    return res.status(400).json({ error: "User ID is required" });
+  }
+
+  if (id == 1) {
```
kylestang, line 22: Shouldn't matter but js type conversion scares me
```suggestion
  if (id === 1) {
```

apps/backend/user/src/login.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  try {
+    const user = await db.oneOrNone<User>(
+      "SELECT * FROM users WHERE email = $1",
+      [email],
+    );
+
+    if (!user) {
+      return res.status(401).json({ error: "Invalid email" });
```
abookout, line 20: resops (Max) might not like this because it allows a malicious agent to sniff out which emails have an account on the site, making it easier to target accounts for cracking their passwords etc

apps/backend/user/src/logout.ts
```diff
@@ -0,0 +1,10 @@
+import { Request, Response } from "express";
+import { IDatabase } from "pg-promise";
+
+const logout = async (req: Request, res: Response, db: IDatabase<object>) => {
```
abookout, line 4: Is this being left for resops to finish?

apps/backend/user/src/patchUser.ts
```diff
@@ -0,0 +1,106 @@
...
+    return res
+      .status(200)
+      .json({ message: "User updated successfully", user: returnObject });
+  } catch (err: any) {
+    if (
+      err?.message.includes("duplicate key value violates unique constraint")
+    ) {
+      return res.status(400).json({ error: "Username already exists" });
+    }
+    return res.status(500).json({ error: (err as Error).message });
```
kylestang, line 102:
```suggestion
    return res.status(500).json({ error: "Something went wrong" });
```

apps/backend/user/src/patchUser.ts
```diff
@@ -0,0 +1,106 @@
...
+    const hashedPassword = password ? await bcrypt.hash(password, 10) : null;
+
+    // Construct patched user!
+    const patchedUser = {
+      username: username || originalUser.username,
+      name: name || originalUser.name,
+      bio: bio || originalUser.bio,
+      profile_pic_url: profilePictureUrl || originalUser.profile_pic_url,
+      verified: verified || originalUser.verified,
+      password: hashedPassword || originalUser.password,
```
abookout, line 61: hashedPassword is not used in the patch

kylestang (changes requested):
Mostly good, a few changes to go through.

Nit: Sometimes the endpoints use `console.log(err)` and other times `console.error(err)`. Would be nice if those were consistent

apps/backend/user/src/login.ts
Scott-Kenning: I'll change both catches to just return "invalid email or password"

apps/backend/user/src/logout.ts
Scott-Kenning: correct

apps/backend/user/src/patchUser.ts
Scott-Kenning: good catch

apps/backend/user/cypress/fixtures/example.json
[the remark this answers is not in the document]
Scott-Kenning: @islandj123

apps/backend/user/src/patchUser.ts
kylestang: Also, the error isn't logged on this endpoint like it is on the other

apps/backend/user/cypress/fixtures/example.json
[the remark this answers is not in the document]
Scott-Kenning: I removed this in the other PR

apps/backend/user/cypress.config.ts
```diff
@@ -0,0 +1,9 @@
+import { defineConfig } from "cypress";
+
+export default defineConfig({
+  e2e: {
+    baseUrl: 'http://localhost:8211',
```
MNThomson, line 5: This being hard-coded makes me unhappy, but not sure there's a better solution

apps/backend/user/cypress.config.ts
```diff
@@ -0,0 +1,9 @@
...
+    specPattern: 'cypress/e2e/**/*.cy.{js,jsx,ts,tsx}',
```
MNThomson, line 6: These aren't e2e tests being written

apps/backend/user/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
MNThomson, line 5: This file?

apps/backend/user/src/createUser.ts
```diff
@@ -0,0 +1,71 @@
...
+import { User } from "./models/user";
+import bcrypt from "bcryptjs";
+
+// createUser route
+const createUser = async (
+  req: Request,
+  res: Response,
+  db: IDatabase<object>,
+) => {
+  const { username, password, email, name, bio, profile_pic_url } = req.body;
```
MNThomson, line 12: I didn't realize all these fields were on the Create user endpoint

apps/backend/user/src/createUser.ts
```diff
@@ -0,0 +1,71 @@
...
+    return res
+      .status(400)
+      .json({ error: "Password does not meet constraints" });
+  }
+
+  // Hash the password
+  const hashedPassword = await bcrypt.hash(password, 10);
+
+  const query = `
+    INSERT INTO users (username, email, password, name, bio, profile_pic_url, verified)
```
MNThomson, line 42: Errrrrrm, hashed password please
```suggestion
    INSERT INTO users (username, email, hashedPassword, name, bio, profile_pic_url, verified)
```

apps/backend/user/src/createUser.ts
```diff
@@ -0,0 +1,71 @@
...
+  `;
+
+  const values = [
+    username,
+    email,
+    hashedPassword,
+    name,
+    bio,
+    profile_pic_url,
+    false,
```
MNThomson, line 54: These values just don't match the sql return statement at all

apps/backend/user/src/deleteUser.ts
```diff
@@ -0,0 +1,64 @@
...
+  req: Request,
+  res: Response,
+  db: IDatabase<object>,
+) => {
+  // This is like this so resops can replace the userid but the guard on id === 1 doesn't show an error, I'm not crazy
+  const id = 1 === 1 ? 2 : 1;
+
+  if (id === 1) {
+    return res.status(401).json({ error: "Cannot delete the base user" });
+  }
```
MNThomson, line 14: Why can't we delete the "base user"? Also, what is the "base user"?

apps/backend/user/src/deleteUser.ts
```diff
@@ -0,0 +1,64 @@
...
+  SET user_id = 1
+  WHERE user_id = $1;
+
+  -- Update listings table buyer_id to deleted user
+  UPDATE listings
+  SET buyer_id = 1
+  WHERE buyer_id = $1;
+
+  COMMIT;
+`;
```
MNThomson, line 39: Is user ID 1 the "deleted user" user? I believe we have in the db a cascade delete so that when a user is deleted, their postings and message conversations are deleted (think GDPR, right to be forgotten, etc)

apps/backend/user/src/deleteUser.ts
```diff
@@ -0,0 +1,64 @@
...
+  RETURNING *;    
+  `;
+
+  try {
+    await db.none(query, [id]);
+
+    await db.oneOrNone(query2, [id]).then((deletedUser) => {
+      if (!deletedUser) {
+        return res.status(404).json({ error: "User not found" });
+      }
```
MNThomson, line 54: Error handling is good but can we get to this case? They have to have a valid JWT but maybe it hasn't expired yet? This seems like a case that the general tey except could handle

apps/backend/user/src/getUser.ts
```diff
@@ -0,0 +1,36 @@
...
+import { IDatabase } from "pg-promise";
+import { User } from "./models/user";
+
+const getUser = async (req: Request, res: Response, db: IDatabase<object>) => {
+  // TODO add JWT auth
+
+  const { id } = req.params;
+
+  const query = `
+        SELECT * FROM users WHERE user_id = $1
```
MNThomson, line 11: Let's request specific properties, using `*` is an antipattern

apps/backend/user/src/login.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  try {
+    const user = await db.oneOrNone<User>(
+      "SELECT * FROM users WHERE email = $1",
+      [email],
+    );
+
+    if (!user) {
+      return res.status(401).json({ error: "Invalid email or password" });
+    }
```
MNThomson, line 21: This exposes us to a timing attack. An attacker can try email/password combos and can check whether or not an email exists in the db by how long the response takes (short if not signed up since it returns early, long if the email exists since then more cryptographic operations are performed on the password).

The easiest way to mitigate this is to always make a bcrypt comparison either against the hashed password in the db or a static string of a random value.

apps/backend/user/src/login.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    if (process.env.NODE_ENV === "test") {
+      isPasswordValid = true;
+    } else {
+      isPasswordValid = await bcrypt.compare(password, user.password);
+    }
+
+    if (!isPasswordValid) {
+      return res.status(401).json({ error: "Invalid email or password" });
+    }
```
MNThomson, line 33:
```suggestion
    if (!isPasswordValid || !iuserExists) {
      return res.status(401).json({ error: "Invalid email or password" });
    }
```

apps/backend/user/src/login.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    const token = "ResopsPleaseReplaceThisWithARealToken";
+
+    res.cookie("token", token, {
+      httpOnly: true,
+      secure: process.env.NODE_ENV === "production",
+      maxAge: 3600000, // 1 hour
+    });
```
MNThomson, line 41: I will be implementing the auth cookie Tue night, let's remove this (since there's a lot of work needed in the JWT area)

apps/backend/user/src/models/user.ts
```diff
@@ -0,0 +1,12 @@
...
+  username: string;
+  email: string;
+  password: string;
+  name: string;
+  bio: string;
+  profile_pic_url: string;
+  verified: boolean;
+  created_at: Date;
+  modified_at: Date;
+}
```
MNThomson, line 12: Not sure if this is used in multiple places, but all the user endpoints have a slightly different return type

apps/backend/user/src/patchUser.ts
```diff
@@ -0,0 +1,107 @@
...
+  req: Request,
+  res: Response,
+  db: IDatabase<object>,
+) => {
+  console.log(req.body);
+
+  // This is like this so resops can replace the userid but the guard on id === 1 doesn't show an error, I'm not crazy
+  const id = 1 === 1 ? 3 : 1;
+
+  const { username, password, name, bio, profilePictureUrl, verified } =
```
MNThomson, line 15: We will need password checks (8 chars, lower/upper, etc) here also

MNThomson, line 15: Also, no way they are allowed to update their own verified status
```suggestion
  const { username, password, name, bio, profilePictureUrl } =
```

apps/backend/user/src/patchUser.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  if (
+    !(
+      username ||
+      password ||
+      name ||
+      bio ||
+      profilePictureUrl ||
+      verified !== undefined
```
MNThomson, line 33:
```suggestion
```

apps/backend/user/src/patchUser.ts
```diff
@@ -0,0 +1,107 @@
...
+        WHERE user_id = $7
+        RETURNING user_id, username, email, name, bio, profile_pic_url, verified, created_at, modified_at;
+      `;
+
+    const updated_user = await db.oneOrNone(patchUserQuery, [
+      patchedUser.username,
+      patchedUser.name,
+      patchedUser.bio,
+      patchedUser.profile_pic_url,
+      patchedUser.verified,
```
MNThomson, line 76:
```suggestion
```

apps/backend/user/src/patchUser.ts
```diff
@@ -0,0 +1,107 @@
...
+      profile_pic_url: profilePictureUrl || originalUser.profile_pic_url,
+      verified: verified || originalUser.verified,
+      password: hashedPassword || originalUser.password,
+    };
+
+    const patchUserQuery = `
+        UPDATE users
+        SET username = $1, name = $2, bio = $3, profile_pic_url = $4, verified = $5, password = $6
+        WHERE user_id = $7
+        RETURNING user_id, username, email, name, bio, profile_pic_url, verified, created_at, modified_at;
```
MNThomson, line 68: This return seems to have stuff that's never used

apps/data/database/populatedb.sql
```diff
@@ -1,9 +1,9 @@
 INSERT INTO users (username, email, password, name, bio, profile_pic_url, verified) VALUES
-('user', 'user@uvic.ca', 'Testing1!', 'User', 'Hi! Im a verified user!', 'https://api.dicebear.com/8.x/adventurer/svg?seed=Jasper', TRUE),
+('deleted_user', 'deleted@martletplace.com', 'iGKT2wOGvPkn5s6U', 'Deleted User', '', 'https://api.dicebear.com/8.x/adventurer/svg?seed=Jasper', TRUE),
```
MNThomson, line 2: Ah, understood. Still confused why we aren't just allowing the cascade delete to do its thing though. Do we really want that stuff staying around?

MNThomson (changes requested):

apps/data/database/populatedb.sql
Scott-Kenning: Yes, we don't want messages or reviews (primarily messages) to be removed when a user is deleted

apps/backend/user/src/createUser.ts
Scott-Kenning: the hashed password is passed in the value, to the field named password

MNThomson: Am idot

apps/backend/user/src/createUser.ts
```diff
@@ -0,0 +1,63 @@
...
+
+// createUser route
+const createUser = async (
+  req: Request,
+  res: Response,
+  db: IDatabase<object>,
+) => {
+  const { username, password, email, name } = req.body;
+
+  if (!username || !password || !email) {
```
MNThomson, line 14: No validation on name?

apps/backend/user/src/createUser.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Hash the password
+  const hashedPassword = await bcrypt.hash(password, 10);
+
+  const query = `
+    INSERT INTO users (username, email, password, name, verified)
+    VALUES ($1, $2, $3, $4, $5, )
+    RETURNING user_id, username, email, name, bio, profile_pic_url;
+  `;
+
+  const values = [username, email, hashedPassword, name, false];
```
MNThomson, line 47: And reason for `false` to not be in the sql query?

```suggestion
  const values = [username, email, hashedPassword, name];
```

apps/backend/user/src/createUser.ts
```diff
@@ -0,0 +1,63 @@
...
+      if (!data) {
+        return res.status(500).json({ error: "User not created" });
+      }
+
+      return res.status(201).send(data);
+    });
+  } catch (err) {
+    console.log(err);
+    return res.status(500).json({ error: "Something went wrong" });
+  }
```
MNThomson, line 60: Nothing required, but why are we combing `await` and `.then`?

apps/backend/user/src/deleteUser.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  COMMIT;
+`;
+
+  const query2 = `    
+-- Delete user
+  DELETE FROM users
+  WHERE user_id = $1
+  RETURNING *;    
+  `;
```
MNThomson, line 46: Future work to combine into one query. If one fails, the entire operation should rollback

apps/backend/user/src/login.ts
kylestang: I considered this and the old issue where the error messages distinguished between wrong username and wrong password. I don't think it matters if people know a username/email is in the DB

apps/backend/user/src/login.ts
```diff
@@ -0,0 +1,51 @@
...
+    );
+
+    let isPasswordValid = false;
+
+    if (process.env.NODE_ENV === "test") {
+      isPasswordValid = true;
+    } else {
+      isPasswordValid = user?.password
+        ? await bcrypt.compare(password, user.password)
+        : false;
```
MNThomson, line 26: This won't run if the user doesn't exist. Still vulnerable to a timing attack.

THIS IS PROBABLY NOT VALID CODE but serves as a reference
```suggestion
      let maybePassword = user.password || "";
      isPasswordValid = await bcrypt.compare(password, maybePassword);
```

MNThomson (approved):
I'm proactively approving this to unblock you if I don't get back to this tonight. Merge it when you feel it's ready

apps/backend/user/src/deleteUser.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  -- Update reviews table to display deleted user
+  UPDATE reviews
+  SET user_id = 1
+  WHERE user_id = $1;
+
+  -- Update listings table buyer_id to deleted user
+  UPDATE listings
+  SET buyer_id = 1
+  WHERE buyer_id = $1;
```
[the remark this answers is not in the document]
Scott-Kenning: this is the buyer ID. for deleted users, the delete will cascade and delete the listings

MNThomson:
LGTM on resolving timing attack

kylestang (approved):
Approving to get this moved along, but I think we should review the user deletion process in sprint 3. There are a bunch of edge cases I'm not sure we have handled

abookout (approved):

PR #208, opened 2024-06-24
Frontend changes for user auth + user context (storing the user)

### LINKED ISSUE #90: User Registration functionality

Scott-Kenning:
# Description
Frontend changes for user auth, + storing user in context

Closes #90

## How to Test
Run, create account, login, navigate to profile page

## Checklist


CaelumD25 (approved):
It looks okay to me, but the tests do need to be passing before merging. As long as that works I'm happy with this

apps/frontend/src/_axios_instance.tsx
```diff
@@ -2,7 +2,7 @@ import axios from "axios";
 
 // Create an instance of axios with custom default configurations
 const _axios_instance = axios.create({
-  baseURL: "https://whateverthiswillbe.com/api/", // Your custom base URL
+  baseURL: "http://localhost/api/", // Your custom base URL
```
MNThomson, line 5: This should be whatever hostname is in the url bar

apps/frontend/src/contexts/UserProvider.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const setUserAndStopLoading = (user: User) => {
+    localStorage.setItem("user", JSON.stringify(user));
+    setUser(user);
+    setLoading(false);
+  };
+
+  const logout = async () => {
+    setUser(null);
+    localStorage.removeItem("user");
+    Cookies.remove("token");
```
MNThomson, line 28: Front-end will not have any access to the token. It's not being sent in the http response body. It's in a `Set-Cookie` header that has `httponly = true` meaning that the front-end CANNOT read (not that you shouldn't, but that the browser will not give it to you)

MNThomson (changes requested):
Not sure why there is anything related to cookies in this pr. The front-end physically (I guess virtually???) cannot access the auth cookie/token

MNThomson (approved):

islandj123 (approved):
tis code

PR #209, opened 2024-06-24
Arun/julian/r198 add recommender training data

### LINKED ISSUE #198: Add recommender training data to database: listings table
## Feature Description
The amazon dataset that the recommender was trained on needs to be added to the backend database so that those items can be properly suggested in the app when a user receives recommendations.

relates: #43

arunalex62:
# Description
The amazon dataset that the recommender was trained on needs to be added to the backend database so that those items can be properly suggested in the app when a user receives recommendations.

Closes #198

## Checklist


Scott-Kenning (approved):

christianbookout (approved):

apps/data/database/Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM postgres:16-alpine
 
 COPY ./postgresql.conf /etc/postgresql.conf
 
+COPY recommender/trainingData.csv /docker-entrypoint-initdb.d/recommender/
```
MNThomson, line 5: Are these items being added to the legitimate database or only for the locally seeded dev instance? This seems like populating the local dev env but not adding to production, so this should probably be added to the docker compose path share and not the Dockerfile

apps/data/database/populatedb.sql
```diff
@@ -5,12 +5,17 @@ INSERT INTO users (username, email, password, name, bio, profile_pic_url, verifi
...
-(5, NULL, 'Listing Five', 'Description for listing five', 500, ROW(51.5074, -0.1278), 'AVAILABLE', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Kiki']);
+-- INSERT INTO listings (seller_id, buyer_id, title, description, price, location, status, image_urls) VALUES
+-- (1, NULL, 'Listing One', 'Description for listing one', 100, ROW(40.7128, -74.0060), 'AVAILABLE', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Jasper', 'https://api.dicebear.com/8.x/bottts/svg?seed=Bella', 'https://api.dicebear.com/8.x/bottts/svg?seed=Bella']),
+-- (2, 3, 'Listing Two', 'Description for listing two', 200, ROW(34.0522, -118.2437), 'SOLD', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Lucy']),
+-- (3, 4, 'Listing Three', 'Description for listing three', 300, ROW(41.8781, -87.6298), 'REMOVED', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Angel']),
+-- (4, NULL, 'Listing Four', 'Description for listing four', 400, ROW(29.7604, -95.3698), 'AVAILABLE', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Casper']),
+-- (5, NULL, 'Listing Five', 'Description for listing five', 500, ROW(51.5074, -0.1278), 'AVAILABLE', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Kiki']);
```
MNThomson, line 13: Uhhhh, any reason to remove these example listing items? They contain as wide a range of different combos as I could find so devs don't have to create items to test their own behaviour

MNThomson (changes requested):

apps/data/database/Dockerfile
kylestang, line 5: Why is it both copying `trainingData.csv` into the dockerfile and creating a volume? Shouldn't we do one or the other?

docker-compose.yml
```diff
@@ -136,7 +136,7 @@ services:
     container_name: martletplace_database
     restart: unless-stopped
     depends_on: []
-    volumes: [ ./apps/data/database/populatedb.sql:/docker-entrypoint-initdb.d/populatedb.sql ]
+    volumes: [ ./apps/data/database/populatedb.sql:/docker-entrypoint-initdb.d/populatedb.sql, ./apps/data/database/recommender/trainingData.csv:/docker-entrypoint-initdb.d/recommender/trainingData.csv ]
```
kylestang, line 139: Same as above

apps/data/database/populatedb.sql
islandj123: Fair enough, I don't have a reason to not keep them.

apps/data/database/Dockerfile
islandj123: I believe we will just be needing this in test, so it is fine to just leave the volume I added for the csv in the docker-compose (as that yml won't be used for production) and remove this redundant copy?

MNThomson: Yep

MNThomson (commented):
LGTMing once the last conversation is resolved

MNThomson (approved):

PR #239, opened 2024-06-26
Add hardcoded authorization parameter to api/recommendations call for demo

arunalex62:
# Description
Hardcode authorization =5 for the recommender demo.

### DIFF
--- a/apps/frontend/src/pages/Homepage.tsx
+++ b/apps/frontend/src/pages/Homepage.tsx
@@ -128,7 +128,7 @@ const Homepage = () => {
       if (initialRender.current) {
         initialRender.current = false;
         _axios_instance
-          .get("/recommendations", { params: { page: 1, limit: 24 } })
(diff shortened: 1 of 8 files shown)


apps/frontend/src/pages/Homepage.tsx
```diff
@@ -128,7 +128,10 @@ const Homepage = () => {
       if (initialRender.current) {
         initialRender.current = false;
         _axios_instance
-          .get("/recommendations", { params: { page: 1, limit: 24 } })
+          // remove hardcoded authorization once it can correctly be handled/after Sprint2 demo.
+          .get("/recommendations", {
+            params: { authorization: 5, page: 1, limit: 24 },
```
sejal-agarwal, line 133: a whole bunch of backend stuff uses auth 1, I'm wondering if we should set it to that.

sejal-agarwal (approved):

PR #241, opened 2024-06-26
[Auth] Implement authentication

### LINKED ISSUE #192: Authentication package
## Feature Description

An authentication package for both JS & Python.

Help the developers by providing a secure by default library for them to use

MNThomson:
# Description

Implement the auth package for use in both python and JS

Closes #192

## Checklist

### DIFF
--- a/apps/backend/lib/src/auth.ts
+++ b/apps/backend/lib/src/auth.ts
@@ -8,7 +8,7 @@ interface UserToken {
 
 export interface AuthenticatedRequest extends Request {
   user: UserToken;
-  // @typescript-eslint/no-explicit-any
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   body: any;
 }
 
@@ -37,10 +37,7 @@ export function authenticate_request(
     "POST /api/user/send-confirmation-email",
     "POST /api/user/confirm-email",
   ];
-  if (
-    unauthenticatedRoutes.includes(path) ||
-    JWT_PUBLIC_KEY === "SKIP_VALIDATION"
-  ) {
+  if (unauthenticatedRoutes.includes(path)) {
     next();
     return;
   }
--- a/apps/backend/lib/src/auth.ts
+++ b/apps/backend/lib/src/auth.ts
@@ -1,5 +1,5 @@
 import { Request, Response, NextFunction } from "express";
-import * as jwt from "jsonwebtoken";
+import { SignOptions, verify, sign } from "jsonwebtoken";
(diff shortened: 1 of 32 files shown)


apps/backend/lib/src/auth.ts
```diff
@@ -0,0 +1,77 @@
+import { Request, Response, NextFunction } from "express";
+import * as jwt from "jsonwebtoken";
+import { Algorithm } from "jsonwebtoken";
```
Scott-Kenning, line 3: Can we clean up this import? Preferabbly just import whats needed from jsonwebtoken

apps/backend/lib/src/auth.ts
```diff
@@ -0,0 +1,77 @@
...
+  userId: number;
+}
+
+export interface AuthenticatedRequest extends Request {
+  user: UserToken;
+}
+
+const SIGNOPTIONS: jwt.SignOptions = {
+  algorithm: "RS256" as Algorithm,
+  expiresIn: "14 days",
```
Scott-Kenning, line 15: I think we agreed on a shorter expiry, but this is probably much nicer for dev

apps/backend/lib/src/auth.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // @ts-ignore
+  let decoded: jwt.JwtPayload & UserToken;
+
+  const authCookie = req.cookies["authorization"];
+  try {
+    decoded = jwt.verify(authCookie, JWT_PUBLIC_KEY, { algorithms: ["RS256"] });
+  } catch (error) {
+    console.error(error);
+    res.status(401);
```
Scott-Kenning, line 38: can we have a json message attached to this with a descriptive error

Scott-Kenning: same goes for the other returns

MNThomson: Api spec says no response body

apps/backend/lib/src/auth.ts
MNThomson: I don't recall discussing an expiration time, but I have no issues changing this in the future

MNThomson:
This PR is feature complete so I'll be passing it off to @UVicMartletplace/backend to get the tests passing (all auth tests are green, but the initial backend tests written are extremely flaky and don't take auth into consideration)

TobySMurray (approved):
LGTM, nice job

Scott-Kenning (approved):

PR #259, opened 2024-06-27
Viewing profiles of other users

### LINKED ISSUE #258: Allow people to view other user's profile
## Feature Description

relates: #xxx
blocked by: #xxx

sejal-agarwal:
# Description

Screenshot 2024-06-27 at 9 54 06 AM

Closes #258 

## Checklist

### DIFF
--- a/apps/frontend/src/pages/ViewProfile.tsx
+++ b/apps/frontend/src/pages/ViewProfile.tsx
@@ -1,4 +1,4 @@
-import { Avatar, Box, TextField, Typography } from "@mui/material";
+import { Avatar, Box, Typography } from "@mui/material";
 import _axios_instance from "../_axios_instance";
 import { useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
@@ -70,10 +70,10 @@ const ViewProfile = () => {
             textAlign: "center",
           }}
         >
-          <Typography variant="h6" gutterBottom>
+          <Typography variant="h6" gutterBottom id="username">
             <strong>Username:</strong> {profile.username}
           </Typography>
-          <Typography variant="h6" gutterBottom>
+          <Typography variant="h6" gutterBottom id="name">
             <strong>Name:</strong> {profile.name}
           </Typography>
           <Box
--- a/apps/frontend/src/pages/ViewProfile.tsx
+++ b/apps/frontend/src/pages/ViewProfile.tsx
@@ -1,0 +1,97 @@
+import { Avatar, Box, TextField, Typography } from "@mui/material";
+import _axios_instance from "../_axios_instance";
+import { useEffect, useState } from "react";
+import { useParams } from "react-router-dom";
+import SearchBar from "../components/searchBar";
+
+interface profileObject {
+  userID: string;
+  username: string;
+  name: string;
+  bio: string;
+  profileUrl: string;
+}
+
+const ViewProfile = () => {
+  const [profile, setProfile] = useState<profileObject>({
+    userID: "",
+    username: "",
+    name: "",
+    bio: "",
+    profileUrl: "",
+  });
+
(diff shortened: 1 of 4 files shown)


Scott-Kenning:
The text field form element implies that the text inside should be an editable field, which it isn't. The bio, name, etc should just be displayed as regular text on the page

It also might be cool to have a "view listings" link which navigates to the search page with their username in the query already, but that is just an idea

apps/frontend/src/pages/ViewProfile.tsx
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  const { id } = useParams<{ id: string }>();
+
+  // Load the profile info for another user from the api given an ID
+  useEffect(() => {
+    console.log(`Fetching data for userID: ${id}`);
+    _axios_instance
+      .get("/user/" + id)
+      .then((response) => {
```
Scott-Kenning, line 31: async / await

apps/frontend/src/pages/ViewProfile.tsx
```diff
@@ -0,0 +1,97 @@
...
+        <Typography variant="h4" sx={{ marginTop: 2 }}>
+          User Profile
+        </Typography>
+        <Avatar
+          src={profile.profileUrl}
+          alt="Profile Picture"
+          sx={{ width: 150, height: 150, mt: 2, mb: 2 }}
+          id="profile_picture"
+        />
+        <Box sx={{ mt: 2, width: "80%", maxWidth: "400px", height: "50px" }}>
```
Scott-Kenning, line 61: this height: 50px is irrelevant since the entire profile text is never going to be less than 50px

apps/frontend/src/pages/ViewProfile.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <TextField
+            label="Bio"
+            variant="outlined"
+            sx={{ width: "100%" }}
+            margin="normal"
+            multiline
+            rows={4}
+            value={profile.bio}
+            id="bio"
+            disabled={true}
```
Scott-Kenning, line 89: These should not be text fields as discussed in the comment

apps/frontend/src/pages/ViewProfile.tsx
sejal-agarwal: Apologies, copy pasta mistake 🫣

apps/frontend/tests/component_tests/ViewProfile.cy.tsx
```diff
@@ -0,0 +1,52 @@
...
+      statusCode: 200,
+      body: expectedProfileObject,
+    }).as("profileRequest");
+
+    // Wait for the mock request to complete and check its status
+    cy.wait("@profileRequest").then((interception) => {
+      // Log request and expected bodies for debugging
+      cy.log("Request Body", interception.request.body);
+      cy.log("Expected Body", expectedProfileObject);
+
```
CaelumD25, line 46: Looks good, but I'd feel better if these were added to one of these tests
```suggestion
cy.contains("Username: " + expectedProfileObject.username).should("be.visible");
cy.contains("Name: " + expectedProfileObject.name).should("be.visible");
cy.contains(expectedProfileObject.bio).should("be.visible");

cy.get("div#profile_picture > img").should("have.attr", "src", expectedProfileObject.profileUrl);
```

CaelumD25 (approved):
Looks good to me, I'd just like a couple more checks added to the tests but other than that it looks good to me!

apps/frontend/tests/component_tests/ViewProfile.cy.tsx
sejal-agarwal: added, thanks!

PR #268, opened 2024-06-29
Added pagination middleware + improved docker config

### LINKED ISSUE #119: Create a Pagination middleware
## Feature Description
Pagination middleware for all API endpoints to use when returning a list of results

We need to avoid duplicating work between the different services to implement pagination

abookout:
# Description

Just the one middleware. All it does is ensure that the `num_items` and `offset` fields are on the req.body so that any endpoints implementing pagination can implement pagination like the following:

```
await db.query("
...
LIMIT $X OFFSET $Y",
[..., req.body.num_items, req.body.offset]);
```

Closes #119 

## Checklist
- [ x ] The code includes tests if relevant
- [ x ] I have *actually* self-reviewed my changes and done QA

### DIFF
--- a/apps/backend/lib/src/pagination.ts
+++ b/apps/backend/lib/src/pagination.ts
@@ -5,7 +5,7 @@ import { Request, Response, NextFunction } from "express";
 export const usePagination = (
   req: Request,
   res: Response,
-  next: NextFunction
+  next: NextFunction,
 ) => {
   const num_items = parseInt(req.query.num_items as string) || 10;
   const offset = parseInt(req.query.offset as string) || 0;
--- a/apps/backend/lib/src/pagination.ts
+++ b/apps/backend/lib/src/pagination.ts
@@ -1,0 +1,15 @@
+import { Request, Response, NextFunction } from "express";
+
+// Pagination middleware. Just ensures that the two pagination parameters are on
+// the request body so you don't need to error check for them in every route.
+export const usePagination = (
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  const num_items = parseInt(req.query.num_items as string) || 10;
+  const offset = parseInt(req.query.offset as string) || 0;
+  req.body.num_items = num_items;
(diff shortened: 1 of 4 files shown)


apps/backend/lib/src/pagination.ts
```diff
@@ -0,0 +1,15 @@
+import { Request, Response, NextFunction } from "express";
```
MNThomson, line 1: If we're just using the express types, why are we importing the full react dependency?

apps/backend/lib/src/pagination.ts
```diff
@@ -0,0 +1,15 @@
...
+
+// Pagination middleware. Just ensures that the two pagination parameters are on
+// the request body so you don't need to error check for them in every route.
+export const usePagination = (
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  const num_items = parseInt(req.query.num_items as string) || 10;
+  const offset = parseInt(req.query.offset as string) || 0;
```
MNThomson, line 11: Can this throw? If so, let's make sure that our error catching Middleware is in front of this so we don't accidently leak stack traces

abookout: Nope, if `parseInt` can't parse out an int, it returns a `NaN` which is falsy, so the expression would resolve to `10` (or `0`).

apps/backend/lib/src/pagination.ts
abookout: Few reasons
1. The intention is to have tests for lib, which will require express (and did when it was implemented). The reason we don't is because we're still in the middle of migrating the test framework
2. It literally doesn't matter because lib isn't built separately from the microservices which use it, and obviously they all use express, so no extra build files are created anyway

Scott-Kenning (approved):

MNThomson (approved):

MNThomson (commented):
Looks like CI needs some love

MNThomson (approved):

PR #275, opened 2024-06-30
Fix search bar bugs

### LINKED ISSUE #274: Fixes to the homepage and searchbar
## Describe the bug

- [x] Fix the defaults for the price filters should be null, rather than zero
- [x] Fix the "Not Recommended" button on recommended listings take you to the listing, rather than greying out the card as expected
- [x] Change the behaviour of updating filters to either apply automatically on search, or through some graphical indicator if that's easier
- [x] And the spinner for low loading

Relates: #272

ChloeZacharias:
# Description

- [X] Fix the defaults for the price filters should be null, rather than zero
- [X] Fix the "Not Recommended" button on recommended listings takes you to the listing, rather than greying out the card as expected
- [X] Change the behavior of updating filters to either apply automatically on search, or through some graphical indicator if that's easier
- [X] Add spinner for loading

Closes #274 

## Checklist

### DIFF
--- a/apps/frontend/src/components/Spinner.tsx
+++ b/apps/frontend/src/components/Spinner.tsx
@@ -1,0 +1,65 @@
+import CircularProgress from "@mui/material/CircularProgress";
+import { Grid, Typography } from "@mui/material";
+import { useEffect, useState } from "react";
+
+interface Props {
+  text: string;
+}
+
+export default function Spinner({ text }: Props) {
+  const [isAlive, setIsAlive] = useState<boolean>(true);
+
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      setIsAlive(false);
+    }, 40000);
+    return () => clearTimeout(timer);
+  }, []);
+
+  return (
+    <>
+      {isAlive ? (
+        <Grid
+          gap={2}
+          container
+          direction="column"
+          alignItems="center"
+          justifyContent="center"
+          sx={{ marginTop: "40px" }}
+        >
+          <Grid item>
+            <CircularProgress size="40px" />
+          </Grid>
+          <Grid item sx={{ textAlign: "center" }}>
+            <Typography variant="h5">Hang with us</Typography>
+            <Typography variant="h6">{text}</Typography>
+          </Grid>
+        </Grid>
+      ) : (
+        <Grid
+          gap={2}
+          container
+          direction="column"
+          alignItems="center"
+          justifyContent="center"
+          sx={{ marginTop: "40px" }}
+        >
+          <Grid item>
+            <CircularProgress
+              variant="determinate"
+              value={100}
+              size="40px"
+              sx={{ color: "red" }}
+            />
+          </Grid>
+          <Grid item sx={{ textAlign: "center" }}>
+            <Typography variant="h5">Network Error</Typography>
+            <Typography variant="h6">
+              Please try again some other time
+            </Typography>
+          </Grid>
+        </Grid>
+      )}
+    </>
+  );
+}
--- a/apps/frontend/src/components/filters.tsx
+++ b/apps/frontend/src/components/filters.tsx
@@ -52,14 +52,14 @@ const Filters = ({ filters, onFilterChange }: FiltersProps) => {
       setMinPrice(null);
     } else if (!regex.test(event.target.value)) {
       setPriceError(
-        "This price is not valid, please make sure the value is positive and in the form xx.xx"
+        "This price is not valid, please make sure the value is positive and in the form xx.xx",
       );
     } else if (newMinPrice === null) {
       setPriceError("");
       setMinPrice(null);
     } else if (maxPrice !== null && newMinPrice > maxPrice) {
       setPriceError(
-        "This price is not valid, please make sure the min price is less than the max price"
+        "This price is not valid, please make sure the min price is less than the max price",
       );
     } else {
       setPriceError("");
@@ -68,8 +68,12 @@ const Filters = ({ filters, onFilterChange }: FiltersProps) => {
     }
     setMinPrice(newMinPrice === null ? null : +newMinPrice);
     onFilterChange({
-      ...filters,
       minPrice: event.target.value === "" ? null : +event.target.value,
+      maxPrice: maxPrice,
+      status: status,
+      searchType: type,
+      latitude: latitude,
+      longitude: longitude,
     });
   };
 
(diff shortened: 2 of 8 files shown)


CaelumD25 (approved):
Looks good to me! Covers all the issues we found with this on Thursday.

apps/frontend/src/components/filters.tsx
```diff
@@ -83,19 +99,42 @@ const Filters = ({ filters, onFilterChange }: FiltersProps) => {
...
+      status: status,
+      searchType: type,
+      latitude: latitude,
+      longitude: longitude,
+    });
   };
 
   const handleStatusChange = (event: SelectChangeEvent<string>) => {
     setStatus(event.target.value);
+    onFilterChange({
```
sejal-agarwal, line 115: Similar to above comment: 

```
onFilterChange({
...filterState,
status: newStatus
});
```

apps/frontend/src/components/filters.tsx
```diff
@@ -83,19 +99,42 @@ const Filters = ({ filters, onFilterChange }: FiltersProps) => {
...
+      status: event.target.value,
+      searchType: type,
+      latitude: latitude,
+      longitude: longitude,
+    });
   };
 
   const handleTypeChange = (event: SelectChangeEvent<string>) => {
     setType(event.target.value);
+    onFilterChange({
```
sejal-agarwal, line 127:
```
onFilterChange({
...filterState,
searchType: newSearchType
});
```

sejal-agarwal (approved):

apps/frontend/src/pages/Homepage.tsx
```diff
@@ -245,6 +222,14 @@ const Homepage = () => {
...
           </>
         )}
+        {!searchCompleted && listingObjects.length === 0 ? (
+          <Spinner text="Loading listings..." />
+        ) : null}
+        {searchCompleted && listingObjects.length === 0 ? (
+          <Typography variant="h6" component="h3" gutterBottom>
+            No Results Found
+          </Typography>
+        ) : null}
```
Scott-Kenning, line 232:
```suggestion
        {listingObjects.length === 0 && (searchCompleted ? ... : ...)}
```

apps/frontend/src/components/Spinner.tsx
```diff
@@ -0,0 +1,65 @@
+import CircularProgress from "@mui/material/CircularProgress";
```
Scott-Kenning, line 1: Is this component actually needed? In theory the loading state should be handled by whatever component is loading. And having the timeout here instead of on the request is also not how this should be done, as its not actually representative of the loading. And if an automatic timeout is being set, 40 seconds is way too long. 3-5 seconds is a more appropriate timeout

apps/frontend/src/components/Spinner.tsx
```diff
@@ -0,0 +1,65 @@
...
+          justifyContent="center"
+          sx={{ marginTop: "40px" }}
+        >
+          <Grid item>
+            <CircularProgress size="40px" />
+          </Grid>
+          <Grid item sx={{ textAlign: "center" }}>
+            <Typography variant="h5">Hang with us</Typography>
+            <Typography variant="h6">{text}</Typography>
+          </Grid>
```
Scott-Kenning, line 36: If you do decide to keep this component, this is the only difference between the 2 states, so you can just have conditionals here on the spinner and the text

apps/frontend/src/components/filters.tsx
```diff
@@ -49,10 +49,14 @@ const Filters = ({ filters, onFilterChange }: FiltersProps) => {
...
     if (event.target.value === "") {
       setPriceError("");
+      setMinPrice(null);
     } else if (!regex.test(event.target.value)) {
       setPriceError(
         "This price is not valid, please make sure the value is positive and in the form xx.xx",
       );
+    } else if (newMinPrice === null) {
+      setPriceError("");
+      setMinPrice(null);
```
Scott-Kenning, line 59: What is the case where newMinPrice is null and not ""?

apps/frontend/src/components/Spinner.tsx
CaelumD25: The search/recommendation endpoint has taken over 30 seconds in the past to respond, so the default timeout on our axios requests is now 40 seconds until timeout. I know that this isn't technically ideal, but having a simple spinner makes the website look better and provides the user with some information other than a blank screen which may seem like a website failure to many users. 

The way it's set up now, it requires minimal effort to add it to our existing pages so I'd like to keep it like this if possible.

apps/frontend/src/components/filters.tsx
ChloeZacharias: Since the number is accepted in a text box when nothing is entered the `event.target.value` is "" but the minPrice can not hold that value so when a search happens and I am setting the value to null to show the empty string. This is why I am checking both.

apps/frontend/src/pages/Homepage.tsx
ChloeZacharias: updated

apps/frontend/src/components/filters.tsx
ChloeZacharias: Changed

apps/frontend/src/components/filters.tsx
ChloeZacharias: Changed

apps/frontend/src/components/filters.tsx
```diff
@@ -83,19 +99,42 @@ const Filters = ({ filters, onFilterChange }: FiltersProps) => {
       const priceValue: number = +event.target.value;
       setMaxPrice(priceValue >= 0 ? priceValue : priceValue * -1);
     }
-    setMaxPrice(event.target.value ? Number(event.target.value) : null);
+    setMaxPrice(newMaxPrice === null ? null : +newMaxPrice);
+    onFilterChange({
+      minPrice: minPrice,
```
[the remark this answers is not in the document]
ChloeZacharias: changed

apps/frontend/src/components/Spinner.tsx
CaelumD25: I tried it a few ways and I thought the way it is now was the simplest to read.

apps/frontend/src/components/filters.tsx
```diff
@@ -83,19 +99,42 @@ const Filters = ({ filters, onFilterChange }: FiltersProps) => {
...
+      maxPrice: event.target.value === "" ? null : +event.target.value,
```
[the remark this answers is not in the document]
ChloeZacharias: This was the easiest for me to read when debugging I might keep it this way since it is not a big change, incase I need to debug more in the future

apps/frontend/src/components/filters.tsx
```diff
@@ -83,19 +99,42 @@ const Filters = ({ filters, onFilterChange }: FiltersProps) => {
       const priceValue: number = +event.target.value;
       setMaxPrice(priceValue >= 0 ? priceValue : priceValue * -1);
     }
-    setMaxPrice(event.target.value ? Number(event.target.value) : null);
+    setMaxPrice(newMaxPrice === null ? null : +newMaxPrice);
```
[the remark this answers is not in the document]
ChloeZacharias: This was the easiest for me to read when debugging I might keep it this way since it is not a big change, incase I need to debug more in the future

apps/frontend/src/components/filters.tsx
ChloeZacharias: I ended up not changing these because the way java seems to be updating it is not getting the correct values when making multiple updates quickly

PR #287, opened 2024-07-03
Gavin/271 pgsync dying on build

### LINKED ISSUE #271: PGSync Dying on build
## Describe the bug
Sometimes, when attempting to build the app with `docker-compose up --build`, PGSync fails to index 2000 or so listings and exits with code 1. It looks as follows:

Image

Sometimes it works, sometimes it doesn't. When it works, it works like a dream (search works with filtering, pagination, etc. and recommender tunes based on the search results all on the frontend):

Image

When it works, I was able to add a listing on the frontend, and it immediately showed up in my searches afterwards, confirming that pagination was working.

## To Reproduce
1. Run `docker-compose down`
2. Run `docker-compose up --build`
3. Look for the following in the logs (this marks a failure):

Image

May require a few tries to reproduce, and may depend on your computer.

## Expected behaviour
This issue doesn't ever happen, or there is a retry mechanism to ensure it eventually works in a few tries

GDeane:
# Description
To fix this issue, the only important change was to update the README (ensuring the user runs `docker-compose down -v` before running `docker-compose up -–build`)

Closes #271

## How to Test
Run through the steps in the search engine readme and nothing should break

## Checklist

### DIFF
--- a/apps/algorithm/search/README.md
+++ b/apps/algorithm/search/README.md
@@ -2,12 +2,11 @@
 
 1. Clone repository.
 
-2. In martletplace directory run `docker-compose down -v` to ensure no containers are running and any cached volumes are deleted.
+2. In martletplace directory run `docker-compose up -–build`. This step may take a while (few minutes to build) as it is
+   as large application.
 
-3. In martletplace directory run `docker-compose up -–build`. This step may take a while (few minutes to build) as it is as large application.
+3. Successful output looks like this at the end (takes a while):
 
-
-4. Successful output looks like this at the end (takes a while):
 ```
 	martletplace_database       | 2024-06-27 03:52:57.947 GMT [36] STATEMENT:  FETCH FORWARD 1 FROM "c_7fa83c326a80_3"
 	martletplace_database       | 2024-06-27 03:52:57.947 GMT [36] LOG:  logical decoding found consistent point at 0/1C0BFD8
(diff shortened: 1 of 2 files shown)


apps/algorithm/search/README.md
```diff
@@ -2,10 +2,12 @@
 
 1. Clone repository.
 
-2. In martletplace directory run `docker-compose –build`. This step may take a while (few minutes to build) as it is as large application.
+2. In martletplace directory run `docker-compose down -v` to ensure no containers are running and any cached volumes are deleted.
```
MNThomson, line 5: This doesn't work. `-v` can only find and delete the anonymous volumes if the stack is still running

GDeane: sadge, will update the readme to create instructions that will work.

emilasuska (approved):
LGTM

MNThomson (approved):

apps/algorithm/search/README.md
```diff
@@ -14,27 +15,32 @@
...
 
-  1. Try sorting functionality by clicking the “Sort By” button.
+8. Try sorting functionality by clicking the “Sort By” button.
 
 ## To run automated tests:
-Open a separate terminal and run `docker exec -it martletplace_search bash`
+
+Once the application is built using `docker-compose up --build -d`, open a separate terminal and run `docker exec -it martletplace_search bash`
 
 In the resulting bash terminal, run `pytest src/test_server.py`
```
christianbookout, line 46: Could condense this into just `docker exec -it martletplace_search pytest src/test_server.py` but obviously not a big deal

christianbookout (approved):

PR #294, opened 2024-07-04
[Frontend] Fix Axios being hardcoded to `localhost`

### LINKED ISSUE #293: Axios `localhost` is hardcoded for api calls
## Describe the bug

Axios defaults to using `localhost` for all api calls. This is obviously bad.

## Expected behaviour

Axios should use whatever hostname is in the url bar

MNThomson:
# Description

Replace the hardcoded `localhost` with the proper `window.location` APIs

Closes #293

## How to Test

Access the frontend from both `localhost` and `local.martletplace.ca` and ensure the api calls don't get blocked by CORS

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,17 @@
 # Clone the repository
 git clone https://github.com/UVicMartletplace/martletplace && cd martletplace
 
-# To start developing, run the docker compose stack
-docker compose up --build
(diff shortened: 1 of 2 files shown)


CaelumD25 (approved):
Looks good to me! Can't see a reason this should cause us any problems on the frontend side

PR #322, opened 2024-07-05
Fix pgsync-search race condition

### LINKED ISSUE #321: Pgsync sometimes fails to start
## Describe the bug

PGsync fails to start if the search service starts first and creates the `martletplace` elasticsearch index before pgsync can. There's a race condition since both depend on elasticsearch: healthy, so the bug only appears sometimes.

## To Reproduce
1. Add a dependency to docker-compose to start pgsync after search (eliminates flakiness)
2. docker compose up --build

## Expected behaviour

Pgsync creates the martletplace index.

kylestang:
# Description

Fixes a race condition where pgsync and the search service both attempt to create the `martletplace` elasticsearch index, which causes pgsync to crash if it loses the race

Closes #321 

## How to Test

`docker compose up --build`

## Checklist

### DIFF
--- a/apps/algorithm/search/src/routes.py
+++ b/apps/algorithm/search/src/routes.py
@@ -13,18 +13,6 @@ search_router = APIRouter()
 
 es = Elasticsearch([ES_ENDPOINT], verify_certs=False)
 
-if not es.indices.exists(index=DEFAULT_INDEX):
-    es.indices.create(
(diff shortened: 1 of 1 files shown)


GDeane (approved):
LGTM, the only thing of note is that if I want to get the location part of search working again later, I'll have to figure out what the correct place to map location as a geo_point is.

PR #324, opened 2024-07-05
Scott/backend237 update email verification

### LINKED ISSUE #237: Email verification on account sign up
## Feature Description

An email with a verification link needs to be sent once the user has signed up

Scott-Kenning:
# Description

Updating email verification to work with auth tokens. In dev, a link is printed to console (and in the datalayer docker container) that needs to be clicked to confirm your email address after signing up

Closes #237 

## How to Test

npm test

## Checklist


apps/backend/user/src/sendComfirmationEmail.ts
```diff
@@ -30,21 +31,22 @@ const sendConfirmationEmail = async (
...
-  const token = "jwttoken";
+  const token = create_token({ userId: userId });
   const body = `
     <p>Please click the link below to confirm your email</p>
     <a href="http://localhost/confirm/${token}"> Confirm Email </a>   
   `;
 
   try {
-    await axios.post("http://localhost/api/email", {
+    await axios.post("/api/email", {
```
kylestang, line 41: This should use the EMAIL_ENDPOINT env var provided by docker_compose

kylestang (changes requested):

sejal-agarwal (approved):
lgtm :) thanks scott!

kylestang (approved):
Approving, but would like those checks moved to top level

apps/backend/user/src/sendConfirmationEmail.ts
```diff
@@ -30,21 +29,22 @@ const sendConfirmationEmail = async (
   }
 
   const subject = "MartletPlace - Please confirm your email";
-  const token = "jwttoken";
+  const token = create_token({ userId: userId });
```
MNThomson, line 32: This should have the optional second argument to restrict it to a specific route. See the auth.ts file in lib

apps/backend/user/src/confirmEmail.ts
```diff
@@ -1,14 +1,42 @@
 import { Request, Response } from "express";
 import { IDatabase } from "pg-promise";
+import { verify, JwtPayload } from "jsonwebtoken";
 
 const confirmEmail = async (
   req: Request,
```
MNThomson, line 6:
```suggestion
   req: AuthenticatedRequest,
```

apps/backend/user/src/confirmEmail.ts
```diff
@@ -1,14 +1,42 @@
...
+    return res.status(401).json({ error: "Invalid code" });
+  }
+
+  // The userId is currently stored incorrectly in the jwt as structured: userId = { user_id: number }
+  // This should be updated at some point
+  const userId = decoded?.userId?.user_id;
+
+  if (!userId) {
+    return res.status(400).json({ error: "Invalid user ID in code" });
+  }
```
MNThomson, line 39: Why are we doing JWT things here? This endpoint should be called from the frontend with the auth cookie (and then the auth Middleware will magically handle this case if you restrict the token to this route only)

apps/frontend/src/pages/ConfirmEmail.tsx
```diff
@@ -10,7 +10,7 @@ const ConfirmEmail = () => {
   useEffect(() => {
     const confirmEmail = async () => {
       try {
-        await axios.post("/user/confirm-email", { code: token });
+        await axios.post("/api/user/confirm-email", { code: token });
```
MNThomson, line 13: If you also make sure the token is set as a cookie, then none of this auth implementation is required

MNThomson (commented):
I'm worried about this auth/jwt stuff. You're currently giving the user a full login token to their email, so it's not exactly "insecure", but that token would allow anyone to login for any endpoint without having to first go to the verify email endpoint (which could really muck things up)

islandj123 (approved):
Looks good, tested and functions as expected

apps/backend/user/src/confirmEmail.ts
Scott-Kenning: Because the confirm email page is coming from the link sent in an email, so a cookie can’t be sent

apps/backend/user/src/sendConfirmationEmail.ts
```diff
@@ -30,21 +29,22 @@ const sendConfirmationEmail = async (
...
   const body = `
     <p>Please click the link below to confirm your email</p>
     <a href="http://localhost/confirm/${token}"> Confirm Email </a>   
   `;
 
   try {
-    await axios.post("http://localhost/api/email", {
+    await axios.post(`${process.env.EMAIL_ENDPOINT}`, {
```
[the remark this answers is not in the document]
Scott-Kenning: Related to both comments about this it’s checked in the lib/auth folder on startup I just have the checks here for peace of mind

PR #326, opened 2024-07-06
Use "lat" and "lon" field names consistently everywhere for geo location, consistent with elasticsearch naming.

### LINKED ISSUE #260: Latitude and Longitude search filtering and sorting broken during pgsync syncing
## Describe the bug
I'm running into the problem where elasticsearch requires location fields to be mapped as geo_point with the fields "lat" and "lon" in order to search by location (https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-point.html). 

previously, (when using reindexing API) location fields looked like this:
`"location": {
            "lat": 45.4215,
            "lon": -75.6972
          },`

Now, they look like this after using PGSync:
`
"location": {
            "latitude": 40.7128,
            "longitude": -74.006
          },
`

When I search by location, I currently get this error:
image.

In short, if location can't be changed to "lat" and "lon" instead of "latitude" and "longitude" using the PGSync schema (or some other method), we won't be able to search within 5 km of or sort by location.

Meanwhile, PGSync has the issue where it does not recognize the LOCATION_TYPE defined in PGSync
image

I have not been able to figure out how to resolve this.

_Originally posted by @GDeane in https://github.com/UVicMartletplace/martletplace/issues/214#issuecomment-2190722842_

## To Reproduce
1. Uncomment the location based filtering and sorting code in the /api/search endpoint in routes.py
Image

3. In terminal in the martletplace directory, run `docker-compose down` , then `docker-compose up --build elasticsearch database pgsync search` (this builds a reduced portion of the application)
4. Once the application has finished building successfully, uncomment this test in test_server.py:
Image
5. In a fresh terminal, run `docker exec -it martletplace_search bash`
6. In the resulting bash terminal, run `pytest src/test_server.py` (this runs the tests)

## Expected behaviour
All tests pass without issue

## Actual (problematic) behavior
A number of tests fail for this reason:
Image

## Suggested fix
Figure out a way to modify the pgsync schema.json to apply the following transformation:

In postgres
`"location": {
            "latitude": 45.4215,
            "longitude": -75.6972
          },`

In elasticsearch
`
"location": {
            "lat": 40.7128,
            "lon": -74.006
          },
`

Everything location related should work then.

emilasuska:
# Description

Use "lat" and "lon" field names consistently everywhere for geo location, consistent with elasticsearch naming.

Also, create the index with the "location" field typed as a "geo_point".

Closes #260 

## How to Test

1. In terminal in the martletplace directory, run `docker-compose down` , then `docker-compose up --build elasticsearch database pgsync search` (this builds a reduced portion of the application)
2. In a fresh terminal, run `docker exec -it martletplace_search bash`
3. In the resulting bash terminal, run `pytest src/test_server.py` (this runs the tests)

## Checklist
- [X] The code includes tests if relevant

### DIFF
--- a/apps/algorithm/search/src/routes.py
+++ b/apps/algorithm/search/src/routes.py
@@ -4,7 +4,7 @@ from typing import Dict, Any
 from elasticsearch import Elasticsearch, NotFoundError
 from fastapi import APIRouter, HTTPException, Header
 
-from .config import DEFAULT_INDEX, ES_ENDPOINT
+from .config import DEFAULT_INDEX, ES_ENDPOINT, DISTANCE_TO_SEARCH_WITHIN
 from .database import insert_user_search
 from .enums import Status, SearchType, Sort
 from .validation import validate_search_params
@@ -29,8 +29,8 @@ if not es.indices.exists(index=DEFAULT_INDEX):
 @search_router.get("/api/search")
 async def search(
     query: str,
-    latitude: float,
-    longitude: float,
+    lat: float,
+    lon: float,
     page: int = 1,
     limit: int = 20,
     minPrice: float = None,
@@ -40,7 +40,7 @@ async def search(
     sort: Sort = "RELEVANCE",
     authorization: str = Header(None),
 ):
-    validate_search_params(latitude, longitude, page, limit, minPrice, maxPrice)
+    validate_search_params(lat, lon, page, limit, minPrice, maxPrice)
 
     INDEX = os.getenv("ES_INDEX", DEFAULT_INDEX)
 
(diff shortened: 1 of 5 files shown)


apps/algorithm/search/src/routes.py
```diff
@@ -29,8 +29,8 @@
 @search_router.get("/api/search")
 async def search(
     query: str,
-    latitude: float,
-    longitude: float,
+    lat: float,
+    lon: float,
```
MNThomson, line 33: Is this api spec compatible?

GDeane: I don't believe so, the lat and lon fields only need to be named that in elasticsearch specifically (not for our search endpoint). This change should be reverted I believe

apps/frontend/src/pages/CreateListing.tsx
GDeane, line 1: I really don't want us to have to touch frontend code unless absolutely necessary. 

I feel like there must be a transform function that would allow PGSync to change the names of the variables when indexing into elasticsearch without having to touch any other team's code. Something like this but for nested object types: https://pgsync.com/tutorial/transform-nodes/

apps/frontend/src/pages/EditListing.tsx
GDeane, line 1: See comment on CreateListing.tsx

apps/frontend/tests/component_tests/CreateListing.cy.tsx
GDeane, line 1: See comment on CreateListing.tsx

GDeane (changes requested):
This doesn't seem to work with the frontend and the rest of the application. 

It works for tests because the test environment is set up in a way that maps location as a geo_point type for an isolated test index and then puts in listings with location objects where the sub-fields are "lat" and "lon" as expected.

When the application is built, PGSync indexes some 2000 or so listings from postgres into our elasticsearch index (the real, not-isolated one). These listings have "latitude" and "longitude" as their names rather than "lat" and "lon", so they do not nicely map to a geo_point field and thus location doesn't work.

This is the part we really need to fix - changing those names in elasticsearch when they are indexed, not necessarily in our code.

GDeane:
We might need some further discussion to clarify what exactly needs changing to fix location searching.

PR #329, opened 2024-07-06
[Images] Create images api

### LINKED ISSUE #19: Create S3 blobstore base project
## Feature Description

Set up the project and docker for s3 blob storage using Min.io

The frontend needs to store pictures and adding that to our database would make its size explode

MNThomson:
# Description

Creates the basic images api

Closes #19 

## How to Test

Try uploading images in the frontend, observe that they work

## Checklist


MNThomson:
@UVicMartletplace/frontend there is something weird in relation to uploading images. The images api works and the frontend correctly PATCHs the image to the `/api/images` endpoint and the image is successfully stored in the datalayer. However, the `PATCH/POST /api/listing` request does not include this image url. I believe it has something (at least on the `EditListing` page) to be related to the setters being run after the image is uploaded. I tried debugging it but am not confident in the frontend side of the codebase.

TLDR: Image is uploaded to datalayer properly and can be access from `/api/images/<UUID>` but the api request to `/api/list` does not include this image url

MNThomson:
Ok, so sending images from the frontend seems to be completely broken. I'm not sure what is going on actually.

I'm testing the image upload service like:
```
$ curl -X POST --data-binary @file.png http://localhost/api/images
{"url":"/api/images/88e6a770-1b57-4156-a362-f5a048796f0f"}
```

I can then navigate to `http://localhost/api/images/88e6a770-1b57-4156-a362-f5a048796f0f` and view the image properly.

I do not know how or what format the frontend is sending but it does not seem to be the raw binary data

MNThomson:
All frontend usage of images seems to not be working (CreateListing, EditListing, EditUser). They are all uploading something but doesn't seem to be the required binary data.

Additionally, the `EditListing` endpoints isn't sending the new image url to the backend, it's just sending back the image urls from the currently set images

apps/data/datalayer/src/image.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Serialize)]
+struct CreateImageResponse {
+    url: String,
+}
+
+pub fn image_router() -> Router {
+    Router::new()
+        .route("/api/images", post(upload_image_handler))
+        .layer(DefaultBodyLimit::disable())
+        .route("/api/images/:id", get(get_image_handler))
```
kylestang, line 25: Can we add the CatchPanic middleware? See #331 for an example

apps/data/datalayer/src/image.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+async fn upload_image_handler(body: Bytes) -> impl IntoResponse {
+    println!("RECIEVED FILE");
+    let uuid = Uuid::new_v4();
+    let dir_path = std::path::Path::new("images");
+    let file_path = dir_path.join(format!("{}", uuid));
+    if !dir_path.exists() {
+        fs::create_dir_all(dir_path).unwrap();
```
kylestang, line 34: Let's use expect with an error message

apps/data/datalayer/src/image.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    let mut file = File::create(file_path).unwrap();
+    file.write_all(&body).unwrap();
```
kylestang, line 38: Ditto

apps/data/datalayer/src/image.rs
```diff
@@ -0,0 +1,59 @@
...
+    let resp = CreateImageResponse {
+        url: format!("/api/images/{}", uuid),
+    };
+    println!("WROTE TO FILE: {:?}, {}", uuid, body.len());
+    (StatusCode::CREATED, Json(resp))
+}
+
+async fn get_image_handler(Path(id): Path<String>) -> impl IntoResponse {
+    let dir_path = std::path::Path::new("images");
+    let file_path = dir_path.join(id);
```
kylestang, line 48: Do we need to consider directory traversal?

MNThomson: Whoooops

PR #331, opened 2024-07-06
Add ability to send emails from the datalayer's email api

### LINKED ISSUE #328: Send emails from datalayer
## Feature Description

Datalayer should be able to send emails in prod

Datalayer doesn't communicate with email server

kylestang:
# Description

Sends emails from the datalayer to the email server

Closes #328 

## How to Test

1. `docker compose up --build`
2. `docker exec -it martletplace_datalayer bash`
3. `curl -X POST -H 'Content-Type: application/json' -d '{"to": "test@example.com", "subject": "Test email", "body": "Example"}' http://localhost:8302/api/email`

## Checklist

### DIFF
--- a/apps/data/datalayer/src/email.rs
+++ b/apps/data/datalayer/src/email.rs
@@ -3,6 +3,7 @@ use std::env;
 use axum::{http::StatusCode, routing::post, Json, Router};
 use lettre::{AsyncSmtpTransport, AsyncTransport, Message, Tokio1Executor};
 use serde::{Deserialize, Serialize};
+use tower_http::catch_panic::CatchPanicLayer;
 
 #[derive(Deserialize, Debug, Serialize)]
 struct Email {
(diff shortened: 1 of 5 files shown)


apps/data/datalayer/src/email.rs
```diff
@@ -1,23 +1,60 @@
...
-    Router::new().route("/api/email", post(email_handler))
+    Router::new()
+        .route("/api/email", post(email_handler))
+        .layer(CatchPanicLayer::new())
 }
 
 async fn email_handler(Json(email): Json<Email>) -> StatusCode {
     println!("Sending Email: {:?}", email);
+
+    if env::var("SEND_EMAILS").unwrap_or_default() == "TRUE" {
```
MNThomson, line 28: Does this default to empty string? Doesn't matter since it's guarded, but curious.

MNThomson (approved):

apps/data/datalayer/src/email.rs
kylestang: Yep. I originally had `unwrap_or(String::new())` but the linter wanted this instead.

PR #342, opened 2024-07-08
Move backend ci tests out of docker

kylestang:
# Description

Closes #XXX

## Checklist

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -68,15 +68,15 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Docker Compose Up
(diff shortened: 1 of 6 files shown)


Scott-Kenning (approved):

MNThomson (approved):
What about email sending tests? Are those not going to try and make external network calls? We could have that route take in a client that could be mocked in tests

kylestang:
They would if you run it with the environment variable `SEND_EMAILS` set to true, which ci doesn't at the moment. If we want to do that then yes mocking it would be the way to go.

MNThomson:
Ah, but isn't the `SEND_EMAILS` var attached to the datalayer? Meaning that no matter what, the user service is going to try and make an http call to the datalayer?

PR #345, opened 2024-07-09
Add health checks to backend services

### LINKED ISSUE #343: User service healthcheck
## Feature Description

Add healthchecks to the user services

relates: #137

We want to ensure services actually start

kylestang:
# Description

Added healthchecks to all 4 backend services

Closes #343 

## How to Test

`docker compose up --build`
`docker ps`

## Checklist

### DIFF
--- a/apps/data/elasticsearch/Dockerfile
+++ b/apps/data/elasticsearch/Dockerfile
@@ -1,5 +1,5 @@
 FROM docker.elastic.co/elasticsearch/elasticsearch:8.13.4
 
-HEALTHCHECK CMD curl -k https://elastic:$ELASTIC_PASSWORD@localhost:8311/_cluster/health
+HEALTHCHECK CMD curl -kf https://elastic:$ELASTIC_PASSWORD@localhost:8311/_cluster/health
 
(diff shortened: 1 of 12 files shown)


apps/data/elasticsearch/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM docker.elastic.co/elasticsearch/elasticsearch:8.13.4
 
-HEALTHCHECK CMD curl -k https://elastic:$ELASTIC_PASSWORD@localhost:8311/_cluster/health
+HEALTHCHECK CMD curl -kf https://elastic:$ELASTIC_PASSWORD@localhost:8311/_cluster/health
```
MNThomson, line 3: ?

MNThomson (approved):
LGTM

apps/data/elasticsearch/Dockerfile
kylestang: Shouldn't really change functionality because of the way elasticsearch's healthcheck works, but the -f flag makes curl requests fail if the response isn't 200, otherwise 400/500 responses would be interpreted as a success. Decided to toss it in while I was updating the other health checks

PR #346, opened 2024-07-09
Add health checks to algo services

### LINKED ISSUE #344: Algo service healthchecks
## Feature Description

Add healthchecks to user service

relates: #137 

We want to ensure the algo services actually start

kylestang:
# Description

Adds healthchecks for algo services. Built on top of #345, merge that first.

Closes #344

## How to Test

`docker compose up --build`
`docker ps`

## Checklist

### DIFF
--- a/apps/algorithm/recommend/src/server.py
+++ b/apps/algorithm/recommend/src/server.py
@@ -21,7 +21,7 @@ recommender = Recommender()
 @app.middleware("http")
 async def authenticate_request(request: Request, call_next):
     # Allow the healthcheck to pass auth
-    if request.url == "http://localhost:8222/.well-known/health":
+    if request.url.path == "/.well-known/health":
         response = await call_next(request)
         return response
 
--- a/apps/algorithm/recommend/src/server.py
+++ b/apps/algorithm/recommend/src/server.py
@@ -1,7 +1,7 @@
 import ast
 import re
 from typing import List
-from fastapi import FastAPI, HTTPException, Depends, Request
+from fastapi import FastAPI, HTTPException, Depends, Request, Response
 import pandas as pd
 from sqlalchemy import insert
 from sqlmodel import select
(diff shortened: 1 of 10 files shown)


apps/algorithm/recommend/src/server.py
```diff
@@ -20,6 +20,11 @@
 
 @app.middleware("http")
 async def authenticate_request(request: Request, call_next):
+    # Allow the healthcheck to pass auth
+    if request.url == "http://localhost:8222/.well-known/health":
```
MNThomson, line 24: Let's not hard-coded the host, just hard-coded the path

apps/algorithm/search/Dockerfile
```diff
@@ -2,9 +2,13 @@ FROM python:3.12-slim
 
 WORKDIR /app
 
+RUN apt update && apt install -y curl
```
MNThomson, line 5: Is there really nothing installed in the python container for making requests? I'm sure you could do a `python -c "import requests......."` or something

apps/algorithm/search/src/server.py
```diff
@@ -9,6 +9,11 @@
 
 @app.middleware("http")
 async def authenticate_request(request: Request, call_next):
+    # Allow the healthcheck to pass auth
+    if request.url == "http://localhost:8221/.well-known/health":
```
MNThomson, line 13: Ditto

MNThomson (approved):
LGTM

apps/algorithm/search/Dockerfile
kylestang: I did consider that, we can do `python -c "import requests; response = requests.get('link'); exit(0) if response.status_code == 200 else exit(1)"`. I went with this mainly because
1. It's consistent with all the other heathchecks
2. What's one more install? It should be cached by docker and we can pretty much forget about it

I get not wanting to install more stuff, so I can use python if it's preferable

MNThomson: Let's use python. As soon as we run `apt` inside a container, build times increase and the image size massively increases (due to the cache & apt lists data now being stored)

PR #394, opened 2024-07-13
implement charities page

### LINKED ISSUE #298: 𝗣𝗮𝗴𝗲: View Charities Page
## Feature Description

This page is for viewing all of the current and past charities, including their sponsors, and start and end dates. 

We need a page to view current and past charities. 

## Related External Resources

[Mock](https://www.figma.com/design/rgiAgBIuZmCKUR2cjopfmy/MartletPlace?node-id=0-1&t=qpIvfKeI0MBCzLgJ-0)

sejal-agarwal:
# Description
Screenshot 2024-07-13 at 9 06 18 AM

Screenshot 2024-07-13 at 9 06 29 AM

Closes #298 

## Checklist

### DIFF
--- a/apps/frontend/src/components/CharityCard.tsx
+++ b/apps/frontend/src/components/CharityCard.tsx
@@ -1,0 +1,122 @@
+import { Card, CardContent, Typography, Grid, Avatar } from "@mui/material";
+import { useStyles } from "../styles/pageStyles";
+import { colors } from "../styles/colors";
+
+interface Organization {
+  name: string;
+  logoUrl: string;
+  donated: number;
+  receiving: boolean;
+}
+
+interface CharityCardProps {
+  id: string;
+  name: string;
+  description: string;
+  startDate: string;
+  endDate: string;
+  imageUrl: string;
+  organizations: Organization[];
+  funds: number;
+  listingsCount: number;
+}
+
+function formatDate(dateString: string) {
+  const date = new Date(dateString);
+  const year = date.getFullYear();
+  const months = [
+    "January",
+    "February",
+    "March",
+    "April",
+    "May",
+    "June",
+    "July",
+    "August",
+    "September",
+    "October",
+    "November",
+    "December",
+  ];
+  const month = months[date.getMonth()];
+  const day = date.getDate();
+  return `${month} ${day}, ${year}`;
+}
+
+const CharityCard = ({
+  name,
+  description,
+  startDate,
+  endDate,
+  organizations,
+  funds,
+  listingsCount,
+}: CharityCardProps) => {
+  const classes = useStyles();
+
+  const formattedStartDate = formatDate(startDate);
+  const formattedEndDate = formatDate(endDate);
+
+  const receivingOrganizations = organizations.filter((org) => org.receiving);
+  const partnerOrganizations = organizations.filter((org) => !org.receiving);
+
+  return (
+    <Card
+      sx={{ ...classes.charityCard, display: "flex", flexDirection: "column" }}
+    >
+      <CardContent sx={{ flexGrow: 1 }}>
+        <Typography
+          variant="h5"
+          component="h2"
+          color={colors.martletplaceNavyBlue}
+          sx={{ fontWeight: "bold" }}
+        >
+          {name}
+        </Typography>
+        <Typography variant="body2" color="textSecondary" gutterBottom>
+          {description}
+        </Typography>
+        <Typography variant="body1">
+          <strong>Total Funds Donated:</strong> ${funds}
+        </Typography>
+        <Typography variant="body1">
+          <strong>Charity Items Sold:</strong> {listingsCount}
+        </Typography>
+        <Typography variant="body2" color="textSecondary">
+          <strong>Start Date:</strong> {formattedStartDate}
+        </Typography>
+        <Typography variant="body2" color="textSecondary">
+          <strong>End Date:</strong> {formattedEndDate}
+        </Typography>
+        <Typography sx={classes.sectionTitle}>
+          Receiving Organizations:
+        </Typography>
+        <Typography variant="body1">
+          {receivingOrganizations.map((org) => org.name).join(", ")}
+        </Typography>
+        <Grid container spacing={1} sx={{ marginTop: "10px" }}>
+          {receivingOrganizations.map((org, index) => (
+            <Grid item key={index}>
+              <Avatar src={org.logoUrl} alt={org.name} sx={classes.orgLogo} />
+            </Grid>
+          ))}
+        </Grid>
+        <Typography sx={classes.sectionTitle} style={{ marginTop: "20px" }}>
+          Partner Organizations:
+        </Typography>
+        <Typography variant="body1">
+          {partnerOrganizations.map((org) => org.name).join(", ")}
+        </Typography>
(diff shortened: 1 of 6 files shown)


apps/frontend/src/components/CharityCard.tsx
```diff
@@ -0,0 +1,122 @@
...
+  description: string;
+  startDate: string;
+  endDate: string;
+  imageUrl: string;
+  organizations: Organization[];
+  funds: number;
+  listingsCount: number;
+}
+
+function formatDate(dateString: string) {
```
Scott-Kenning, line 24: JS probably has a date .toString() function to format dates, if the date comes in as a parsable date object. Would need a bit of research

apps/frontend/src/pages/Charities.tsx
```diff
@@ -0,0 +1,258 @@
...
+        console.error("Error fetching charities:", error);
+        setLoading(false);
+      }
+    };
+    fetchData();
+  }, []);
+
+  const currentCharity = Array.isArray(charities)
+    ? charities.find((charity) => new Date(charity.endDate) >= currentDate)
+    : undefined;
```
Scott-Kenning, line 59: null should be used here over undefined. Undefined is for values that have not been set yet, whereas null is for values that are intentionally empty

apps/frontend/src/pages/Charities.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+  const pastCharities = Array.isArray(charities)
+    ? charities.filter((charity) => new Date(charity.endDate) < currentDate)
+    : [];
```
Scott-Kenning, line 63: null can be used here instead of a blank array, then you can just check if pastCharities is null or not. That would require an additional check on the pastCharities array though, so this implementation is probably fine too

apps/frontend/src/pages/Charities.tsx
```diff
@@ -0,0 +1,258 @@
...
+        } else {
+          const days = Math.floor(distance / (1000 * 60 * 60 * 24));
+          const hours = Math.floor(
+            (distance % (1000 * 60 * 60 * 24)) / (1000 * 60 * 60),
+          );
+          const minutes = Math.floor(
+            (distance % (1000 * 60 * 60)) / (1000 * 60),
+          );
+          const seconds = Math.floor((distance % (1000 * 60)) / 1000);
+          setTimeRemaining(`${days}d ${hours}h ${minutes}m ${seconds}s`);
```
Scott-Kenning, line 84: Again, looking into JS date to string formatting might make this a lot cleaner

apps/frontend/src/pages/Charities.tsx
```diff
@@ -0,0 +1,258 @@
...
+            gutterBottom
+            sx={{ paddingLeft: "20px", paddingTop: "20px" }}
+          >
+            No Past Charities
+          </Typography>
+        )
+      ) : (
+        <>
+          {currentCharity ? (
+            <>
```
Scott-Kenning, line 151: This fragment is unnecessary. Also it might be nice to factor one or two of these elements out in to componenets to avoid 250+ line files

apps/frontend/src/components/CharityCard.tsx
sejal-agarwal: Screenshot 2024-07-13 at 1 43 55 PM

I tried for quite some time this morning to use a library, but I keep getting the above error when I try and use a library.

I did run `npm install date-fns`

apps/frontend/src/pages/Charities.tsx
sejal-agarwal: I'm going to leave it then

apps/frontend/src/pages/Charities.tsx
sejal-agarwal: Since there's math involved in calculating the live countdown number, I don't think there's much simplification I can do 🤔

apps/frontend/src/components/CurrentCharity.tsx
```diff
@@ -0,0 +1,127 @@
...
+      } else {
+        const days = Math.floor(distance / (1000 * 60 * 60 * 24));
+        const hours = Math.floor(
+          (distance % (1000 * 60 * 60 * 24)) / (1000 * 60 * 60),
+        );
+        const minutes = Math.floor((distance % (1000 * 60 * 60)) / (1000 * 60));
+        const seconds = Math.floor((distance % (1000 * 60)) / 1000);
+        return `${days}d ${hours}h ${minutes}m ${seconds}s`;
+      }
+    };
```
CaelumD25, line 30: This is a really cool functionality! However, it does take a bit to render even on my beefy desktop computer, and it re-renders every time the toggle is pressed, it doesn't personally bother me, but if you're feeling keen to work on it, useCallback might be of some help

image
image

apps/frontend/src/components/CurrentCharity.tsx
```diff
@@ -0,0 +1,127 @@
...
+        {charity.name}
+      </Typography>
+      <Typography
+        variant="body1"
+        sx={{ marginBottom: "20px", textAlign: "center" }}
+      >
+        {charity.description}
+      </Typography>
+      <Box
+        component={Paper}
```
CaelumD25, line 65: This is something that's been at the back of my mind for quite a while, but I'm not sure when/if we should use paper...
For me, it's the exact same as Card except you get less rounded edges, and that's what's used for nearly all the other components that have a height/depth with drop shadows if we're trying to stay consistent. I'd love to get your feedback on this because I'd really like to use Paper more, but I'm just not sure which scenarios it works best in

apps/frontend/src/pages/Charities.tsx
```diff
@@ -0,0 +1,119 @@
...
+            variant="h6"
+            component="h4"
+            gutterBottom
+            sx={{ paddingLeft: "20px", paddingTop: "20px" }}
+          >
+            No Past Charities
+          </Typography>
+        )
+      ) : (
+        currentCharity && <CurrentCharity charity={currentCharity} />
```
CaelumD25, line 113: I think it'd be good to add this to the current charity toggle, that way the user can see the state at a glace and it's consistent with the other page
```suggestion
        <Box>
          <Typography
            variant="h5"
            component="h3"
            gutterBottom
            sx={{ paddingLeft: "20px", paddingTop: "20px" }}
          >
            Current Charity Events
          </Typography>
          {currentCharity && <CurrentCharity charity={currentCharity} />}
        </Box>
```
_You'll need to add the box import_

CaelumD25 (approved):
Looks amazing! Excellent work doing all that, I have a few minor issues but other than that, I don't see anything that explicitly needs to be changed!

Quality Metrics
- [x] Duplicate API calls
- [x] Form validation errors
- [x] Error handling from APIs

apps/frontend/src/components/CurrentCharity.tsx
sejal-agarwal: I think I've been using cards when there are multiple cards being used on a page. No other reason lol.

I opted for paper because I liked the look and wanted to have a nice box to highlight the current charity. 

Let me know if you'd prefer something else :) Not super attached to anything

CaelumD25: Nah, I'm good, I think I was feeling more motivated when I wrote this... Our best judgement is probably fine for this project

PR #402, opened 2024-07-15
Create the Create Charities Page

### LINKED ISSUE #304: 𝗣𝗮𝗴𝗲: Create Create Charities Page
## Feature Description

This page is for creating a new charity, if it is done like this, we may need to handle who can create charities.

This page will be very similar to the other created pages.

relates: #298

CaelumD25:
# Description
This PR adds the route, and page, and mocks the endpoints for creating a listing. This page is for testing charities and isn't intended for regular users(unless we're otherwise told). Therefore, I've kept things like console logs and added buttons so that the state of the charity object can be known by whoever is using this page.

image
image

image

Closes #304 

## How to Test
Go to [localhost/charity/new](http://localhost/charity/new), or run `npm test` or run `npx cypress open` from within the frontend directory

## Checklist


apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,671 @@
...
+  FormControlLabel,
+  Grid,
+  Paper,
+  TextField,
+  Typography,
+} from "@mui/material";
+import SearchBar from "../components/searchBar";
+import { colors } from "../styles/colors";
+import _axios_instance from "../_axios_instance";
+
```
sejal-agarwal, line 19: We keep defining these Organization/Charity objects. We could either implement it in one place like types.ts or my PR that I just merged has them defined in Charities.tsx which you should be able to use from there. It has the `export` keyword so you should be able to use it in your files. 

I do use the Charities object in several files and use the exported type from Charities.tsx

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,671 @@
...
+interface CharityObject {
+  name: string;
+  description: string;
+  startDate: string;
+  endDate: string;
+  imageUrl: string;
+  organizations: OrganizationObject[];
+}
+
```
sejal-agarwal, line 35: I expect these Image objects are also being used in the create listing page too 🤔 Perhaps we could move it more a central location and use it there. Just to make the code more DRY but not a necessity

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,671 @@
...
+    submissionEvent,
+  ) => {
+    submissionEvent.preventDefault();
+    if (sent) return;
+    const logoURL = await asyncListingImageWrapperLogo();
+    if (!logoURL) {
+      alert("Image failed to upload");
+      return;
+    }
+    console.log("LOGO URL", logoURL);
```
sejal-agarwal, line 79: remove console log

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,671 @@
...
+    const newDataObject = await asyncListingImageWrapperPartners(logoURL);
+    if (newDataObject) {
+      console.log("Sending", newDataObject);
```
sejal-agarwal, line 82: I think this console log should be removed too

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,671 @@
...
+          [key]: value,
+        };
+      } catch (error) {
+        const newOrganization: OrganizationObject = {
+          name: "",
+          logoUrl: "",
+          donated: 0,
+          receiving: false,
+        };
+        console.log(updatedOrganizations);
```
sejal-agarwal, line 250: remove

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,671 @@
...
+        updatedOrganizations.push(newOrganization);
+        console.log("Added", updatedOrganizations);
```
sejal-agarwal, line 252: remove

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,671 @@
...
+                    Upload Picture
+                    <input
+                      id="upload-logo-input"
+                      type="file"
+                      hidden
+                      accept="image/*"
+                      onChange={updateCharityImage}
+                    />
+                  </Button>
+                  <TextField
```
sejal-agarwal, line 478: How we're not assigning an end date?

sejal-agarwal (commented):
Looks awesome Caelum! Love the button to add an additional org.

If you can, would love to remove as many console logs as possible. Just a bit nervous about seeing any kind of data in the console. TAs might not like it.

Also, would it be possible to only show the `Donated` input box if it is an org not receiving funds from the charity? Form might be slightly confusing since the donated amount for receiving orgs doesn't mean anything. Again, this is an internal page and I'm not sure the TAs will see it so it might be okay to leave it

apps/frontend/src/pages/CreateCharity.tsx
CaelumD25: Actually, these are all new for creating charities(At least the image upload ones are), since I need to keep track of the order of images here to attach them to the appropriate organization, whereas with the regular create listing page it's not so important.

apps/frontend/src/pages/CreateCharity.tsx
CaelumD25: We definitely could, but aren't they supposed to be only a month long? I had it in here originally, then thought that maybe I don't need to have it since I can just add 31 days to the start date, would love to hear your thoughts though!

apps/frontend/src/pages/CreateCharity.tsx
CaelumD25: Those charity objects are now in the types file!

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+      } catch (error) {
+        alert("Charity Creation Failed");
+        console.error(error);
+        setSent(false);
+      }
+    } else {
+      alert("Images failed to upload, please try again later");
+      setSent(false);
+    }
+  };
```
Scott-Kenning, line 87: Rather than using a "sent" variable to track the status of the request, it might be better to clear all the form fields once a successful request has gone through. Having another variable just adds room for more bugs. I think a setSent(false) is missed if there is no logoURL

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+      }
+
+      // Upload partnerImages
+      const partnerImagesArray = partnerImages
+        .filter((obj) => obj.image !== undefined)
+        .map((obj, index) => ({ image: obj.image as File, index }));
+
+      const partnerImagesObjectArray =
+        await asyncUploadImages(partnerImagesArray);
+      if (partnerImagesObjectArray && partnerImagesObjectArray.length > 0) {
```
Scott-Kenning, line 179:
```suggestion
      if (partnerImagesObjectArray?.length > 0) {
```

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+        const newOrganizations = [...newCharityObject.organizations];
+        for (let i = 0; i < newCharityObject.organizations.length; i++) {
+          newOrganizations[partnerImagesObjectArray[i].index].logoUrl =
+            partnerImagesObjectArray[i].url;
+        }
```
Scott-Kenning, line 184: In practice JavaScript doesn't change the order of an array on its own (as far as I know / have encountered), but the docs specifically state that **array order is not guaranteed**. An implementation that creates organizations one by one rather than dealing with an array of images and indexing them all might be safer and also more simple, avoiding these for loops with fancy indexing (presumably `partherImagesObjectArray[i].index` should just be i?)

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+    updateNewCharityPayload("description", event.target.value);
+  };
+
+  const updateStartDate = (event: ChangeEvent<HTMLInputElement>) => {
+    const dateValue = event.target.value;
+    setStartDate(dateValue);
+    if (dateValue) {
+      const dateStart = new Date(dateValue);
+      const dateEnd = new Date(dateStart);
+      dateEnd.setDate(dateStart.getDate() + 31);
```
Scott-Kenning, line 256: This won't work on months with less than 31 days, or if we wanted to just create 12 charities off the bat for a full year. We should probably be setting the end date manually as well (this page should never be exposed to a user anyways, its more for us to add charities)

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+
+  // Removes an organization, and re-indexes the images
+  const removeOrganization = (index: number) => {
+    setNewCharityObject((prevState) => {
+      const updatedOrganizations = [...prevState.organizations];
+
+      updatedOrganizations.splice(index, 1);
+
+      for (let i = 0; i < partnerImages.length; i++) {
+        console.log("index", i, partnerImages);
```
Scott-Kenning, line 323: console.log should be removed

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+    let i = 0;
+    for (; i < newCharityObject.organizations.length; i++) {
+      if (newCharityObject.organizations[i].name == key) {
+        return i;
+      }
+    }
+    return -1;
+  };
+
+  // Removes an organization, and re-indexes the images
```
Scott-Kenning, line 315: All of this logic could be removed if images were always attached to their organization

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+  const removeOrganization = (index: number) => {
+    setNewCharityObject((prevState) => {
+      const updatedOrganizations = [...prevState.organizations];
+
+      updatedOrganizations.splice(index, 1);
+
+      for (let i = 0; i < partnerImages.length; i++) {
+        console.log("index", i, partnerImages);
+        if (partnerImages[i].index == index) {
+          const updatedPartnerImages = partnerImages;
```
Scott-Kenning, line 325: This should be a let instead of const if its being updated

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+                    multiline
+                    label="Charity Description"
+                    id="charity-description"
+                    onChange={updateDescription}
+                    value={newCharityObject.description}
+                  />
+                  <img
+                    src={logoImageString}
+                    style={{
+                      display: logoImageString ? "block" : "none",
```
Scott-Kenning, line 436: We should probably be conditionally displaying this element based on logoImageString rather than setting it to have a css display property of "none". Less html to load and in theory a user could remove the none property and then see the image element (second reason is probably not that relevant). Same pattern is used a few times

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+                      maxHeight: "300px",
+                      width: "auto",
+                      aspectRatio: "auto",
+                      objectFit: "contain",
+                    }}
+                    alt={"Charity Event Logo"}
```
Scott-Kenning, line 442:
```suggestion
                    alt="Charity Event Logo"
```

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+                    {newCharityObject.organizations.map(
+                      (organization, index) => (
+                        <Box key={index} sx={{ marginBottom: "10px" }}>
+                          <Grid container>
+                            <Grid item lg={6} sm={12}>
+                              <Box sx={{ margin: "10px" }}>
+                                <TextField
+                                  type="text"
+                                  label="Organization Title"
+                                  id={"org-title-" + index}
```
Scott-Kenning, line 486: All of these `"id-" + index` can be changed to `id-${index}`

apps/frontend/src/pages/CreateCharity.tsx
```diff
@@ -0,0 +1,651 @@
...
+                      fontSize: "16px",
+                      padding: "10px 20px",
+                      margin: "10px",
+                    }}
+                    onClick={() => {
+                      alert(
+                        "The Following Object Will be Submitted(Files are uploaded at the end)\n" +
+                          JSON.stringify(newCharityObject, null, 2),
+                      );
+                      console.log("STATUS\n", newCharityObject);
```
Scott-Kenning, line 636: remove console.log

apps/frontend/src/pages/CreateCharity.tsx
CaelumD25: Yeah, I fiddled with a couple ways to do this, but basically the organizations all need to be attached to a specific image, and the requests aren't guarenteed to be resolved in order, so I need to attach them to the index of the organization that called them. I tried to attach them directly to the organizations when submitting but the object never updated when I needed it to and this was working

apps/frontend/src/pages/CreateCharity.tsx
CaelumD25: Yeah, it didn't work earlier(attaching the image urls to the orgs was far more confusing that I thought), it's possible I could have done something else, but I'm pretty sure this should work...

apps/frontend/src/pages/CreateCharity.tsx
CaelumD25: Fixed

apps/frontend/src/pages/CreateCharity.tsx
CaelumD25: Re-added the end date and added form validation to address this so you guys can create whatever you'd like

apps/frontend/src/pages/CreateCharity.tsx
CaelumD25: Tried this, but for whatever reason it doesn't actually render the image, it just renders the base64 string instead so I think I'll leave it alone so I don't spend too much time on it

apps/frontend/src/pages/CreateCharity.tsx
CaelumD25: Linter doesn't like since I have that object as also possibly boolean

apps/frontend/src/pages/CreateCharity.tsx
CaelumD25: Seemed Scott wanted it so I just added that field back

ChloeZacharias (approved):
Looks great

PR #404, opened 2024-07-16
Chloe/search history

### LINKED ISSUE #299: Add Search History to Search Bar
## Feature Description

Search history needs to be implemented in this sprint, we'll be doing it in a more minimal way than having an actual history management page. The history will include the 5 most recent items and offer a delete button for each(double check spec that this is a requirement).

relates: #71 

Our users need a way to view their search history and manage it.

## Related External Resources

Image

ChloeZacharias:
# Description
Adds a search history to the search bar and a clear search history on the profile page
image
image

Closes #299 

## Checklist

### DIFF
--- a/apps/frontend/src/components/searchBar.tsx
+++ b/apps/frontend/src/components/searchBar.tsx
@@ -19,7 +19,6 @@ import React from "react";
 import useUser from "../hooks/useUser";
 import { colors } from "../styles/colors";
 import _axios_instance from "../_axios_instance.tsx";
-import { Token } from "@mui/icons-material";
 
 interface SearchObject {
   query: string;
--- a/apps/frontend/src/components/searchBar.tsx
+++ b/apps/frontend/src/components/searchBar.tsx
@@ -69,12 +69,7 @@ const SearchBar = () => {
   const [showFilters, setShowFilters] = useState(false);
   const [searchInput, setSearchInput] = useState("");
   const [searchFocus, setSearchFocus] = useState(false);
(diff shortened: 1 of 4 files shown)


apps/frontend/src/components/searchBar.tsx
```diff
@@ -122,6 +132,15 @@ const SearchBar = () => {
       page: 1,
       limit: 8,
     };
+    //NEED TO ADD THE TOKEN TO THE REQUEST
```
Scott-Kenning, line 135: the token is in the request by default as it is a cookie in the browser (if you're logged in)

apps/frontend/src/components/searchBar.tsx
```diff
@@ -190,6 +209,10 @@ const SearchBar = () => {
   const isDesktop = useMediaQuery("(min-width:1240px)");
   const isSmallPage = useMediaQuery("(min-width:740px)");
```
Scott-Kenning, line 210: we should be using the theme breakpoints here instead of manually defining them

apps/frontend/src/components/searchBar.tsx
ChloeZacharias: removed the comment

apps/frontend/src/components/searchBar.tsx
ChloeZacharias: I kept this the same because the other methods I tried were not as responsive `window.innerWidth` `screen.width` and `screen.availWidth`

CaelumD25 (approved):
Looks great! I imagine that there could be some issues with integration but that's a future us problem!

PR #419, opened 2024-07-16
Fix flaky recommender test

### LINKED ISSUE #310: Fix flaky test with recommender
## Describe the bug
Currently, the recommender test is flaky. We should look into fixing it.

christianbookout:
# Description

The recommender tests were flaky after an attempted fix. I believe that fix didn't work since, when we retry loading the model when it gets corrupted, we don't remove the models first. This PR cleans up the code surrounding this issue to make it more transparent, and removes the models before redownloading the model. 

Closes #310 

## Checklist

### DIFF
--- a/apps/algorithm/recommend/src/recommender.py
+++ b/apps/algorithm/recommend/src/recommender.py
@@ -33,21 +33,21 @@ class Recommender:
             self.download_file(RECOMMENDER_URL + url, TRAINING_DIR_URL + url)
 
     def load_model(self, retries=5):
-        if retries < 0:
-            raise Exception("Failed to download the recommender model after a number of retries. There may be a problem with your internet connection, or perhaps you're very unlucky (and should try again).")
-        os.makedirs("/app/src/training", exist_ok=True)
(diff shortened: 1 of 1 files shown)


arunalex62 (approved):

apps/algorithm/recommend/src/recommender.py
```diff
@@ -4,6 +4,13 @@
 import requests
 import os
 
+RECOMMENDER_VERSION = "1.0.1"
+RECOMMENDER_URL = f"https://github.com/UVicMartletplace/martletplace/releases/download/recommender-v{RECOMMENDER_VERSION}/"
+PROCESSED_DATA_URL = "processed_data.csv"
+COSINE_URL = "cosine_similarity_matrix.npy"
+ITEM_VECTORS_URL = "normalized_item_vectors.npy"
+TRAINING_DIR_URL = "/app/src/training/"
+
```
christianbookout, line 13: Perhaps these should be environment variables but this was supposed to be a temporary solution anyway so it doesn't matter.

PR #421, opened 2024-07-16
[Search] Run SQL in background to 3x reponse time

MNThomson:
# Description
The SQL INSERT does not need to happen while the client is waiting. The query now executes in a [FastAPI background task](https://fastapi.tiangolo.com/tutorial/background-tasks/). Any errors resulting from the query will not cause the search to fail but is not concerning as missing the data for a single search history is outweighed by the performance gains.

closes: #422

## How to Test
| Before| After|
|---|---|
|image|image|

## Checklist


MNThomson:
When combined with #425, gets response time down to ~14ms
image

PR #428, opened 2024-07-16
Brent/backend354 not found status fix for backend service

### LINKED ISSUE #354: status 404 should not be returned for "items not found"
These should almost definitely all be 200, if not then probably 400

Image

bmachado18:
# Description
Replaces the 404 error codes with 200 error codes for backend services

Closes #354 

## How to Test
npm run test

## Checklist


Scott-Kenning:
Sorry to be a pain, but could we use 200 instead of 204 to simplify things? There's a good chance frontend will expect a 200 over a 204

Scott-Kenning (approved):

PR #430, opened 2024-07-16
No more recommender in actions testing workflow

christianbookout:
# Description

The recommender is not feasible to test on Actions due to the massive model it needs to download. Perhaps in the future we can implement some caching or something, but right now the recommender tests are flaky and failing at times.


MNThomson:
Fixed by: #495

PR #442, opened 2024-07-21
[Algorithm] Add production container defintions

### LINKED ISSUE #406: Create production docker image for Py services
Use Docker multi stage builds to have both a `dev` and `prod` target. Have docker compose [point to the `dev` target](https://stackoverflow.com/questions/53093487/multi-stage-build-in-docker-compose) (with a global variable so that we can change one line to switch docker compose to target prod).

MNThomson:
# Description
Create productionized containers for Py services

Closes #406

## Checklist

### DIFF
--- a/apps/algorithm/recommend/Dockerfile
+++ b/apps/algorithm/recommend/Dockerfile
@@ -1,4 +1,29 @@
-FROM python:3.12-slim
+FROM python:3.12-slim AS dev
+
+WORKDIR /app
+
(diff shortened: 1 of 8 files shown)


MNThomson:
@kylestang I was messing around with load testing and quickly setup Gunicorn to run the services. Thought I'd commit it so you wouldn't have to replicate the work for working on #406

kylestang (approved):

PR #450, opened 2024-07-21
[Loadtest] Add transaction ramp to relieve pressure on user service

### LINKED ISSUE #449: The loadtest is bursting all requests to the user service at once
## Describe the bug

Loadtest is hitting the user service with all users simultaneously on launch. Meaning it gets absolutely swamped and then kills metrics for everything else (please explain why my P50 on logging in is 40s.....)

MNThomson:
# Description
The "startup time" is somewhat fake as it resets all transactions/scenarios when the real test starts running. Meaning that the user service gets absolutely pounded by every goose at once. Using the test_plan lets us start the transactions slowly and ramp up (with no sudden restarts). This change got RPS +30% and the P50,95 on User -80%

Closes #449

## Checklist


test/load/create_perf_branch.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+git stash --include-untracked
+CURRENT_HEAD=$(git rev-parse HEAD)
+
+git merge max/speed-up-search max/search-database-connection resops/productionize-py-services max/optimize-db-indexes max/increase-pgbouncer-conn-limit max/lower-backend-logging max/test-remove-triggers
```
kylestang, line 7: We should probably make a more general solution before merging to master

MNThomson: Yeah, I can just not commit this file also. Or potentially rename the branches to be `perf/*`

PR #487, opened 2024-07-24
Added charity endpoints

### LINKED ISSUE #432: Create Charities API
## Feature Description

Create charities api which will include logic to calculate funds and associated listings

No charities API

bmachado18:
# Description
Adding charity endpoints

Closes #432 

## How to Test
1. ```npm run test```

## Checklist


Scott-Kenning:
In theory all of these charity endpoints should work, but this PR is not ready for a merge - I think something is missing on the proxy side as the frontend can't seem to make requests to the charity service. The code that adds the listings from the training data csv is also temporarily commented out. This needs to be updated as well to have a charityID field in all the listings before it can be uncommented. Calling it for the night since its 3am

docker-compose.yml
```diff
@@ -49,7 +49,7 @@ services:
     environment:
       <<: *shared_env_var
       EMAIL_ENDPOINT: http://datalayer:8302/api/email
-      SKIP_USER_VERIFICATION: "FALSE"
+      SKIP_USER_VERIFICATION: "TRUE"
```
MNThomson, line 52: This should not be committed

apps/proxy/dynamic_config.yml
```diff
@@ -25,6 +25,10 @@ http:
       rule: "PathPrefix(`/api/search`)"
       service: search
       priority: 200
+    to-charity:
+      rule: "PathPrefix(`/api/charity`)"
+      service: charity
+      priority: 200
```
MNThomson, line 31: You are setting the path to be `/api/charity` yet it seems your endpoints defined in `index.ts` are going to `/api/charities`. Which is it?

islandj123 (approved):
Backend stuff looks good, don't want to hold up resops tomorrow
Time to 😴

GDeane:
The conflict is purely changing the locations of the default 5 listings. To resolve it just use your changes and change the locations of the default five listings to match master.

kylestang (changes requested):
Can we get these fixed? Also, can we get tests for the endpoints other than `createCharities` too?

apps/backend/charity/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+COPY package*.json ./
+COPY tsconfig.json ./
+RUN npm install
+
+COPY --from=lib src /app/lib/src
+COPY src/ ./src
+
+RUN npm run build
+
+HEALTHCHECK CMD curl -f http://localhost:8225/.well-known/health
```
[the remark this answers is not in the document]
Scott-Kenning: will change but fyi this was copy pasted from another service that I think was on port 8223

apps/backend/charity/src/getCurrentCharity.ts
```diff
@@ -0,0 +1,49 @@
...
+        ),
+      ]);
+
+      charityResult.funds =
+        parseFloat(donationFunds.donation_funds) +
+        parseFloat(listingStats.listing_funds);
+      charityResult.listingsCount = parseInt(listingStats.listings_count);
+    }
+
+    res.json(charityResult);
```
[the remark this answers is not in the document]
Scott-Kenning: this is correct

apps/data/database/initdb.sql
```diff
@@ -22,15 +22,26 @@ CREATE TABLE users (
...
 );
 
 CREATE TRIGGER users_modified_at BEFORE UPDATE ON users FOR EACH ROW EXECUTE PROCEDURE trigger_update_modified();
 
+CREATE TABLE charities (
+    charity_id SERIAL PRIMARY KEY,
+    name VARCHAR NOT NULL,
+    description TEXT,
+    start_date DATE,
+    end_date DATE,
```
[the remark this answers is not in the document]
Scott-Kenning: No, charities should be required to have start and end dates

apps/backend/charity/src/getCurrentCharity.ts
```diff
@@ -0,0 +1,49 @@
...
+) => {
+  try {
+    const charityResult = await db.oneOrNone(
+      `SELECT c.charity_id AS id, c.name, c.description, c.start_date AS "startDate", c.end_date AS "endDate", c.image_url AS "imageUrl", 
+        jsonb_agg(jsonb_build_object('name', o.name, 'logoUrl', o.logo_url, 'donated', o.donated, 'receiving', o.receiving)) AS organizations
+       FROM charities c
+       LEFT JOIN organizations o ON c.charity_id = o.charity_id
+       WHERE NOW() BETWEEN c.start_date AND c.end_date
+       GROUP BY c.charity_id;`,
+    );
```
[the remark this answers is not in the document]
Scott-Kenning: Since charity creation is not actually a user function we can assume that there won't be any overlapping, since we are the ones adding the data

apps/backend/charity/src/index.ts
```diff
@@ -0,0 +1,62 @@
...
+import { createCharity } from "./createCharity";
+import { getCharities } from "./getCharities";
+import { getCurrentCharity } from "./getCurrentCharity";
+import { deleteCharity } from "./deleteCharity";
+
+const PORT = 8215;
+
+const app = express();
+
+app.use(morgan("dev"));
```
MNThomson, line 17:
```suggestion
app.use(
  morgan("dev", {
    skip: function (req: Request, res: Response) {
      return res.statusCode < 400;
    },
  }),
);
```
As per the other backend services

apps/backend/charity/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@types/eslint__js": "^8.42.3",
+    "@types/express": "^4.17.21",
+    "@types/morgan": "^1.9.9",
+    "@types/node": "^20.12.13",
+    "@typescript-eslint/eslint-plugin": "^7.13.1",
+    "@typescript-eslint/parser": "^7.13.1",
+    "eslint": "^8.57.0",
+    "eslint-config-prettier": "^9.1.0",
+    "nodemon": "^3.1.2",
+    "pg-promise": "^11.8.0",
```
MNThomson, line 27: This is definitely a run time dependency

apps/backend/charity/src/createCharity.ts
```diff
@@ -0,0 +1,100 @@
...
+      );
+    });
+
+    const results = await db.tx((t) => {
+      return t.batch(queries);
+    });
+
+    const createdOrganizations: Organization[] = results.filter(
+      (result) => result !== null,
+    ) as Organization[];
```
MNThomson, line 76: Would a null result mean an error has occured? In which case shouldn't we return an error (or such)?
 Why else would there be a null?

apps/data/database/initdb.sql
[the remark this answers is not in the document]
MNThomson: So then these should be?
```suggestion
    start_date DATE NOT NULL,
    end_date DATE NOT NULL,
```

apps/data/database/populatedb.sql
```diff
@@ -1,17 +1,31 @@
...
-(5, NULL, 'Listing Five', 'Description for listing five', 500, ROW(48.4805652, -123.3234215), 'AVAILABLE', ARRAY['https://api.dicebear.com/8.x/bottts/svg?seed=Kiki']);
+INSERT INTO charities(name, description, start_date, end_date, image_url) VALUES
+('Charity One', 'A nice clever description for a drab default charity', '2024-01-01', '2027-12-31', 'https://api.dicebear.com/8.x/avataaars/svg?seed=CharityOne'),
+('ScottMagic', 'We will make sure Scott never forgets his Magic cards again', '2022-01-01', '2022-06-14', 'https://api.dicebear.com/8.x/avataaars/svg?seed=CharityTwo'),
+('Bikes4Kids', 'Buying little bicycles for poor little children', '2022-06-15', '2022-06-18', 'https://api.dicebear.com/8.x/avataaars/svg?seed=CharityThree'),
+('Little Treat', 'Fundraise so Anika can afford to have a little treat :-) <3', '2022-06-19', '2022-12-31', 'https://api.dicebear.com/8.x/avataaars/svg?seed=CharityFour'),
+('Evil Charity', 'We are going to do evil stuff muahahaha', '2023-01-01', '2023-12-31', 'https://api.dicebear.com/8.x/avataaars/svg?seed=CharityFive');
```
MNThomson, line 14: Could we change these? I like that they're funny but we probably don't want some of these for demo purposes

apps/data/database/populatedb.sql
```diff
@@ -1,17 +1,31 @@
...
+
+INSERT INTO organizations (name, logo_url, donated, receiving, charity_id) VALUES
+('JPetStore', 'https://api.dicebear.com/8.x/avataaars/svg?seed=OrganizationOne', 100, FALSE, 1),
+('Costco', 'https://api.dicebear.com/8.x/avataaars/svg?seed=OrganizationTwo', 419.68, TRUE, 1),
+('Moms Against Seedless Watermelons', 'https://api.dicebear.com/8.x/avataaars/svg?seed=OrganizationFour', 0.01, TRUE, 2),
+('Seedless Watermelons Against Moms', 'https://api.dicebear.com/8.x/avataaars/svg?seed=OrganizationFive', 12345678.49, FALSE, 4);
```
MNThomson, line 20: Ditto

apps/backend/charity/src/getCurrentCharity.ts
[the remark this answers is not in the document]
MNThomson: Isn't the charity creation API publicly accessible though?

MNThomson (changes requested):
Looks like the new charities endpoint are "correct" according to the OpenAPI schema
```bash
npx @stoplight/prism-cli proxy openapi.yaml http://localhost:80 --port 8000
```

Should every user be able to create/delete orgs/charities? That seems wild and a massive security problem. How should this be handled?

apps/backend/charity/src/createCharity.ts
```diff
@@ -0,0 +1,100 @@
...
+          org.donated,
+          org.receiving,
+          createdCharity.charity_id,
+        ],
+      );
+    });
+
+    const results = await db.tx((t) => {
+      return t.batch(queries);
+    });
```
kylestang, line 72: This won't execute the queries on a single connection. Note that in [the documentation](https://vitaly-t.github.io/pg-promise/Database.html#task) the queries are done within the callback and not on the main database. So essentially move the map inside the callback and replace `db` with `t`.

Nit: from a db performance perspective I'd prefer `task` instead of `tx` since I don't think we need the safety guarantees of a transaction.

apps/backend/charity/src/getCurrentCharity.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    if (!charityResult) {
+      return res.status(200).json({ message: "No active charity" });
```
kylestang, line 21: Nit: is 200 the best response status, or would something else (maybe 404) be better?

kylestang (commented):
All my previous comments have either been addressed or Max picked them up, so the only thing stopping me from approving is the createCharity transaction.

apps/data/database/populatedb.sql
abookout: Go for it

apps/backend/charity/src/getCurrentCharity.ts
abookout: There doesn't need to be an active charity 100% of the time, not having an active charity is expected behaviour therefore 200

apps/data/database/initdb.sql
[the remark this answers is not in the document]
abookout: resolved

abookout:
@MNThomson 
> Isn't the charity creation API publicly accessible though?

I'm not sure your point. Obviously it should not be public in a realistic situation but implementing a way to give only some people access (like an admin user type) is not in the scope of this project.

abookout:
I've changed the behaviour of getCurrentCharity so if there are multiple current charities, the one with the start date closest to the present (i.e. most recent) is chosen

kylestang (approved):
Thanks for working through those!

MNThomson (approved):

PR #500, opened 2024-07-25
Brent/backend286 specific return queries

### LINKED ISSUE #286: FIX RETURNING * with necessary column names
## Feature Description
Some endpoints that access the db use RETURNING * at the end of the query to return all column names resulting from the query. This is bad because the database might change and could be a security issue.

Replace all instances of RETURNING * with the exact column names necessary.

bmachado18:
# Description
In the POST and PATCH queries for listings and reviews, specified which data attributes should be returned explicitely rather than RETURNING *.

Closes #286 
## How to Test
```npm test```

## Checklist


islandj123 (approved):
:godmode:

PR #502, opened 2024-07-25
Check that the user is part of a listing when creating/getting messages

### LINKED ISSUE #429: Improve messaging security
## Feature Description

- Add checks wherever relevant to make sure that message endpoints are only accessible to the current user (e.g. something like sender_id === user_id)

abookout:
# Description
Minor security improvement: ensures the current user is part of the message thread they're trying to interact with

Closes #429


Scott-Kenning (approved):

MNThomson (commented):
Great catch, thanks Anika!
From a performance standpoints, multiple db queries are extremely bad. Curious if there's a way to only have one query per endpoint

PR #515, opened 2024-07-29
Fixed Backend Errors when Creating and Editing Listings 

### LINKED ISSUE #505: Fix Sprint 2 Issues
## Describe the bug

Issues found:
- For empty my listings 404 http code is returned, not handled properly on FE. 
- Create listings form:
  Shows validation error in title field on page load      without touching the form.
  Validation works only for title field, If i fill only the title field and keep other fields empty the form is submitted and the api request is sent with error returned as 400,however it is not shown to the user.
- Responsiveness and general UI can be improved on messaging page

Overall very good job product feels very high quality and most edge cases are covered.

CaelumD25:
# Description
Last sprint there were a couple issues with creating/editing a listing which can be seen in the following along with the fixes I did.

I made this PR since backend seems quite busy, and hopefully this is an easy/non-contentious fix.

- **Description is required to create a listing**
  - Since empty strings are false, change the verification of a request to ensure that description is a string
- **A non-zero price is required for a listing**
  - Since 0 is false, changed the verification of a request to ensure a price is not negative
- **Listing prices get rounded to nearest whole number**
  - In the database, price was stored as an integer which means that nothing can have a fractional price, fixed by changing int to Decimal(10,2)

Closes #505 

## How to Test
Create and/or edit a listing with the above edge cases.

## Checklist

### DIFF
--- a/apps/data/database/initdb.sql
+++ b/apps/data/database/initdb.sql
@@ -32,7 +32,7 @@ CREATE TABLE listings (
     seller_id INTEGER NOT NULL REFERENCES users(user_id) ON DELETE CASCADE,
     buyer_id INTEGER REFERENCES users(user_id),
     title VARCHAR NOT NULL,
-    price REAL NOT NULL,
+    price DECIMAL(10,2) NOT NULL,
     location LOCATION_TYPE NOT NULL,
     status STATUS_TYPE NOT NULL,
     description VARCHAR,
--- a/apps/data/database/initdb.sql
+++ b/apps/data/database/initdb.sql
@@ -32,7 +32,7 @@ CREATE TABLE listings (
     seller_id INTEGER NOT NULL REFERENCES users(user_id) ON DELETE CASCADE,
     buyer_id INTEGER REFERENCES users(user_id),
     title VARCHAR NOT NULL,
-    price INTEGER NOT NULL,
+    price REAL NOT NULL,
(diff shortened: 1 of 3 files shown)


apps/data/database/initdb.sql
```diff
@@ -32,7 +32,7 @@ CREATE TABLE listings (
     seller_id INTEGER NOT NULL REFERENCES users(user_id) ON DELETE CASCADE,
     buyer_id INTEGER REFERENCES users(user_id),
     title VARCHAR NOT NULL,
-    price INTEGER NOT NULL,
+    price REAL NOT NULL,
```
abookout, line 35: We're currently using `DECIMAL(10,2)` in organizations to represent a money amount, this is much preferrable to a floating point format since floats are so lossy
```suggestion
    price DECIMAL(10,2) NOT NULL,
```
`DECIMAL(10,2)` is not perfect either (the ideal would probably be to store an exact bigint of cents) but we're not using real money so it's not a big deal if we're a cent or two off

abookout (changes requested):

apps/data/database/initdb.sql
CaelumD25: Sounds good to me, as long as the user is getting back what they're putting in I'm happy!

abookout (approved):

TobySMurray (approved):

Issue #352, opened 2024-07-10
PATCH and DELETE Review Auth


bmachado18:
## Feature Description
Reviews that are deleted or updated should be verified they own the listing

## Problem You're Trying to Solve
As of now, any verified user can delete and update ANY review. this should be updated

PR #545, opened 2024-07-30
userId check on patch/delete routes

### LINKED ISSUE #351: PATCH and DELETE Listing Auth
## Feature Description
Listings that are deleted or updated should be verified they own the listing

As of now, any verified user can delete and update ANY listing. this should be updated

Scott-Kenning:
Makes it so that users can only delete and modify their own listings and reviews
# Description

Closes #351 #352 

## How to Test
Try and delete another users listing

## Checklist


abookout (approved):
LGTM

TobySMurray (approved):

